PR #20, opened 2025-01-16
Fix setup issue of flagcx backend with the usage of gloo adaptor for torch plugin

MC952-arch:


aoyulong (approved):
LGTM. We should use a systematic naming convention.

PR #22, opened 2025-01-19
Support _allgather_base operation in torch plugin

MC952-arch:


aoyulong (commented):
We need to find a better method to distinguish between different chips rather than relying on macros.

aoyulong (approved):
LGTM

PR #25, opened 2025-01-22
Fix the problem that one request may be missed in extreme cases

haobaba5353:


flagcx/core/proxy.cc
```diff
@@ -132,7 +132,9 @@ inline void* flagcxProxyProgress(void *proxyState_){
   // flagcxSetDevice(proxyState->cudaDev);
   deviceAdaptor->setDevice(proxyState->cudaDev);
 
-  while(!proxyState->progressState.stop || !commplete){
+  int stop = 0;
```
MC952-arch, line 135: commplete -> complete

flagcx/core/topo.h
```diff
@@ -249,4 +249,14 @@ static int mirrorBits(int val, int pow2) {
   for (int b=1, mb=(pow2>>1); b<pow2; b<<=1, mb>>=1) if (val & b) mirror |= mb;
   return mirror;
 }
+
+
+#ifdef CREATE_DEVICE_TOPO_API
```
MC952-arch, line 254: What's this macro for?

MC952-arch (approved):

PR #26, opened 2025-01-24
Fix the AllReduce hang issue in torch plugin

MC952-arch:


flagcx/adaptor/gloo_adaptor.cc
```diff
@@ -34,11 +34,18 @@ flagcxResult_t glooAdaptorCommInitRank(flagcxInnerComm_t *comm, int nranks, flag
...
     try {
-        char line[1024];
-        FLAGCXCHECK(getHostName(line, 1024, '.'));
-        std::string hostname(line);
         ::gloo::transport::tcp::attr attr;
-        attr.hostname = hostname;
+        char line[1024];
+        const char* glooIface = flagcxGetEnv("FLAGCX_GLOO_SOCKET_IFNAME");
+        if(glooIface == NULL) {
+            FLAGCXCHECK(getHostName(line, 1024, '.'));
```
heavyrain-lzy, line 41: Does the machine need to support IPoIB to use IB hardware communication?

heavyrain-lzy (approved):
LGTM

aoyulong (approved):
LGTM. We should design a better mechanism for timing and tracing.

PR #28, opened 2025-02-06
Enable pytorch plugin Coalescing

yzhang35:


plugin/torch/src/backend_flagcx.cpp
```diff
@@ -1,5 +1,6 @@
 #include "backend_flagcx.hpp"
 #include <iostream>
+#include <c10/cuda/impl/CUDAGuardImpl.h>
```
MC952-arch: Better to create a new header file device_flagcx.hpp to handle implementations on different platforms, like event_flagcx.hpp. We need to keep BackendFlagCX well abstracted.

aoyulong:
Please provide a description of the PR.

MC952-arch (approved):
LGTM

PR #32, opened 2025-02-07
Add several minor fixes

MC952-arch:
* Invoke event block call in WorkFlagCX::wait func
* Add plugin/torch/run_hetero.sh script for heterogeneous test
* Reuse sync stream as a substitution of barrier op for better performance

### DIFF
--- a/plugin/torch/example.py
+++ b/plugin/torch/example.py
@@ -27,14 +27,14 @@ print(f"ddp backend config is {dist.get_backend_config()}")
 
 # Create two groups
 ranks = list(range(world_size))
-FLAGCX_GROUP1 = dist.new_group(ranks=ranks, backend=f"cpu:gloo,{dev_name}:flagcx")
-FLAGCX_GROUP2 = dist.new_group(ranks=ranks, backend=f"cpu:gloo,{dev_name}:flagcx")
+FLAGCX_GROUP1 = dist.new_group(ranks=ranks, backend=f"{dev_name}:flagcx")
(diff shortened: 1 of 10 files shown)


plugin/torch/example.py
aoyulong, line 1: Please refactor this file in a more modular way in the future.

plugin/torch/src/backend_flagcx.cpp
```diff
@@ -138,14 +138,11 @@ namespace c10d
...
-        // {
-        //     event_->block(device_id_);
-        // }
-        // if (isBarrierOp_)
-        // For now, we directly call stream sync.
+        if (!coalesced_)
+        {
+            event_->block(device_id_);
+        }
+        if (isBarrierOp_)
```
aoyulong, line 144: What is the reason for this code working now?

plugin/torch/example.py
MC952-arch: OK

plugin/torch/src/backend_flagcx.cpp
MC952-arch: > What is the reason for this code working now?

It seems that a sync stream call is required after flagcxHeteroSend/Recv ops. Such synchronization is added in all C2C algorithms for the moment. Here we adopt the same way as ProcessGroupNCCL does, blocking current stream on the flagcx stream.

aoyulong (approved):
Something will be Improved in the future.

PR #43, opened 2025-02-12
Add naive support for heterogeneous reduce operation and its corresponding op in torch plugin

MC952-arch:
- Implement naive C2C reduce algorithm
- Support reduce op in torch plugin
- Fix bugs about gather and scatter ops in torch plugin


flagcx/flagcx.cc
aoyulong, line 1: It's better to separate the implementations into two levels: one for the heterogeneous clusters and the other for the nodes within a cluster.

MC952-arch: > It's better to separate the implementations into two levels: one for the heterogeneous clusters and the other for the nodes within a cluster.

The C2C part can be encapsulated into a CCLAdaptor, such as CoreCCL or HeteroCCL.

heavyrain-lzy (approved):
LGTM

PR #45, opened 2025-02-13
Add flagcxIsHomoComm() for communicator type verification

MC952-arch:
- Add flagcxIsHomoComm() for communicator type verification
- Do not call group semantics for coalesced collective ops (allreduce, allgather, reducescatter) under heterogeneous circumstance, since current implementations of C2C ops require the use of stream synchronization
截屏2025-02-13 14 30 54

### DIFF
--- a/flagcx/flagcx.cc
+++ b/flagcx/flagcx.cc
@@ -135,6 +135,19 @@ flagcxResult_t flagcxHandleFree(flagcxHandlerGroup_t handler)
     return flagcxSuccess;
 }
 
+flagcxResult_t flagcxIsHomoComm(int *isHomo)
+{
+    if (is_homo_comm())
+    {
+        *isHomo = 1;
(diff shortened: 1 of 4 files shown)


flagcx/flagcx.cc
```diff
@@ -135,6 +135,20 @@ flagcxResult_t flagcxHandleFree(flagcxHandlerGroup_t handler)
     return flagcxSuccess;
 }
 
+
+flagcxResult_t flagcxIsHomoComm(int *isHomo)
+{
```
heavyrain-lzy, line 139: The return value of this function is redundant, it is always `flagcxSuccess`.

MC952-arch: Just to maintain consistent API formatting

heavyrain-lzy (approved):
LGTM

PR #48, opened 2025-02-14
Support async malloc/free ops for flagcxDeviceAdaptor

MC952-arch:


aoyulong (approved):
We need to select a clang-format style for our repository and use it to format all the code at once.

PR #49, opened 2025-02-14
for uninitialized case, flagcx returns error instead of asserting.

yzhang35:


flagcx/flagcx.cc
```diff
@@ -137,6 +137,7 @@ flagcxResult_t flagcxHandleFree(flagcxHandlerGroup_t handler)
 
 flagcxResult_t flagcxIsHomoComm(int *isHomo)
 {
+    FLAGCXCHECK(flagCXEnsureCommReady());
```
MC952-arch, line 140: flagCXEnsureCommReady -> flagcxEnsureCommReady

PR #50, opened 2025-02-17
Deprecate the static communicator cur_comm to prevent state overwriting when instantiating multiple communicators

MC952-arch:


flagcx/flagcx.cc
```diff
@@ -182,24 +174,17 @@ flagcxResult_t flagcxGetUniqueId(flagcxUniqueId_t *uniqueId)
...
     return "Not implemented.";
 }
 
 const char *flagcxGetLastError(flagcxComm_t comm)
 {
-    if (flagCXEnsureCommReady() != flagcxSuccess)
-    {
-        return "Undefined: Comm is not fully initialized.";
+    // TODO: implement a method to retrieve last error string
+    if (comm == NULL) {
```
heavyrain-lzy, line 184: It's better to keep the code style of `{` consistent.

heavyrain-lzy (approved):
LGTM

PR #62, opened 2025-03-03
Add mulit-stream support in torch plugin to parallelize hetero coalesced ops

MC952-arch:
- Each enqueued op for coalescing will be launched on a different stream in heterogeneous mode
- Add stream syncs at the beginning/end of flagcxHeteroSend/Recv ops to ensure correctness


mikethegoblin:
LGTM

mikethegoblin (approved):
LGTM

PR #65, opened 2025-03-06
remove unnecessary work for batch_isend_irecv call

yzhang35:


plugin/torch/src/backend_flagcx.cpp
```diff
@@ -888,15 +888,19 @@ c10::intrusive_ptr<Work> flagcxBackend::send(std::vector<at::Tensor> &tensors,
...
-  std::vector<at::Device> devices{tensor.device()};
-  work->future_ = c10::make_intrusive<c10::ivalue::Future>(
-      c10::ListType::create(c10::TensorType::get()), devices);
-  work->future_->markCompleted(c10::IValue(tensors));
-  return work;
+  if (activeGroupCounter_ <= 0) {
+    // not coalesced
+    work->event_->record(stream, deviceId_);
+    work->deviceId_ = deviceId_;
+    work->coalesced_ = false;
```
MC952-arch: In this way, there is no need to hold work->coalesced_.

MC952-arch (approved):
LGTM

PR #68, opened 2025-03-07
add adaptor functions for topology detection

mikethegoblin:
- added 2 functions in device adaptor for topology detection
- added function to obtain net device properties in `net_ib.h`

### DIFF
--- a/flagcx/include/flagcx.h
+++ b/flagcx/include/flagcx.h
@@ -79,14 +79,6 @@ typedef struct flagcxComm *flagcxComm_t;
 /* Opaque handle to flagcxStream */
 typedef struct flagcxStream *flagcxStream_t;
 
-struct flagcxDevProps {
-  char name[256];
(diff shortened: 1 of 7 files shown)


flagcx/include/flagcx.h
```diff
@@ -79,6 +79,14 @@ typedef struct flagcxComm *flagcxComm_t;
 /* Opaque handle to flagcxStream */
 typedef struct flagcxStream *flagcxStream_t;
 
+struct flagcxDevProps {
```
MC952-arch: maybe move this part into topo.h.

mikethegoblin: I moved the definition to adaptor.h to avoid circular dependency

mikethegoblin: done

MC952-arch (approved):
LGTM

PR #74, opened 2025-03-13
remove gdrSupported field from flagcxDevProps to avoid compilation error

mikethegoblin:
- removed `gdrSupported` from flagcxDevProps to avoid compilation error cuased by different definitions of `cudaDeviceProp` in different cuda versions


MC952-arch (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #83, opened 2025-03-26
convert xml topo to server topo

mikethegoblin:
- added functions that convert xml topo structure to server topo structure

**NOTE**: this is a basic version because we can't fully detect all topology info now. More info will be added to server topo as we add features in topology component.
The following file shows and example of printed server topo
[server_topo.txt](https://github.com/user-attachments/files/19464422/server_topo.txt)


Makefile
```diff
@@ -112,6 +112,7 @@ OBJDIR := $(BUILDDIR)/obj
 INCLUDEDIR := \
 	$(abspath flagcx/include) \
 	$(abspath flagcx/core) \
+	$(abspath flagcx/core/include) \
```
MC952-arch: 这里不加include，h和cc文件都放到core下

MC952-arch (approved):
LGTM

aoyulong (approved):
LGTM

PR #87, opened 2025-03-28
Use host comm to perform reduce operations when comm->has_single_rank_homo_comm is True

MC952-arch:


mikethegoblin (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #110, opened 2025-04-16
gather topology info from all servers

mikethegoblin:
- functions that flatten topoServer for transfer
- function that gathers topology info from all servers, which is needed to construct interserver route


MC952-arch (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #113, opened 2025-04-21
add inter-rank collecting mechanism to initialize the homo inter comm

MC952-arch:
This PR introduces an inter-rank collection mechanism to initialize the homogeneous inter-communicator, composed of a subset of ranks within the homogeneous ranks that are closest to available NICs. This communicator will be used for inter-cluster send/receive operations.


mikethegoblin (approved):
LGTM

PR #115, opened 2025-04-23
read interserver route info from xml file

mikethegoblin:
- read interserver route infor from xml file
- calculate bandwidth of interserver route based on switch and nic bandwidth info
- included rapidxml parser for fast xml parsing


MC952-arch (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #123, opened 2025-04-28
implementation of flagcxGetNicDistance

mikethegoblin:
- added implementation of `flagcxGetNicDistance`
- moved interrank computation into a conditional branch. We only enable interrank computation in hetero mode


MC952-arch (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #125, opened 2025-04-29
add corner case check (inter ranks > totalCount_) and set debug level to TRACE for flagcxPlanner workflow

MC952-arch:
This PR introduces several modifications in flagcxPlanner workflow:
- set single-nic mode if there is a cluster containing N inter ranks superior to totalCount_
- set debug level to TRACE


mikethegoblin (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #128, opened 2025-04-30
change GetNicDistance logic to avoid APUs sharing a nic

mikethegoblin:
- modified logic for getting nic distance for each rank, avoiding the scenario in which multiple gpus share the same nic


heavyrain-lzy (approved):
LGTM

PR #132, opened 2025-05-06
use guid instead of netDev in flagcxNicDistance struct

mikethegoblin:
This PR introduces a fix to potential bug when comparing nic distances to get clusterInterRankList. A nic's dev index is only unique within server scope, not in global scope; hence, it is not an appropriate identifier to use in deduplication.


MC952-arch (approved):
LGTM

PR #137, opened 2025-05-09
Fix heteroFunc setup issue in multi-loop execution with varying inter-rank counts across clusters

MC952-arch:
This PR fix heteroFunc setup issue in multi-loop execution with varying inter-rank counts across clusters


mikethegoblin (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #148, opened 2025-05-14
Fix comparison bug: use sendCount_ >= recvCount_ instead of sendCount_ > recvCount_

MC952-arch:
Otherwise, it will lead to incorrect result for C2C AllReduce operation


mikethegoblin (approved):
LGTM

PR #152, opened 2025-05-15
Replace homoRanks_ with comm_->cluster_sizes[i] in flagcxC2cPlanner::refresh

MC952-arch:
Otherwise, it will lead to incorrect result for C2C AllGather operation


heavyrain-lzy (approved):
LGTM

mikethegoblin (approved):
LGTM

PR #156, opened 2025-05-19
Move AlltoAll/v ops into the flagcxPlanner workflow

MC952-arch:
This PR integrates AlltoAll and AlltoAllv operations into the flagcxPlanner workflow. Instead of using the searchHeteroSendRecvOps function, each rank is treated as an inter-rank to construct customized hetero send/recv strategies. Note that this approach is functional but may not be the most optimized solution for AlltoAll/v operations.


heavyrain-lzy (approved):
LGTM

mikethegoblin (approved):
LGTM

PR #169, opened 2025-05-30
Support cluster split functionality

MC952-arch:
This PR introduces support for splitting a homogeneous cluster into multiple sub-clusters. For instance, given two clusters c1 and c2 with 8 and 16 homo-ranks respectively, setting:

`FLAGCX_CLUSTER_SPLIT_LIST=2,4`

will split:
- c1 into two sub-clusters (c1-1, c1-2) with 4 homo-ranks each
- c2 into four sub-clusters (c2-1, c2-2, c2-3, c2-4) with 4 homo-ranks each

This enables more flexible groupings of ranks for communication planning and facilitates heterogeneous communication scenarios even on a single machine.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -80,7 +80,4 @@ jobs:
       - name: Run unit tests with mpirun
         run: |
           cd /__w/FlagCX/FlagCX/test/unittest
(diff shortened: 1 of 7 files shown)


heavyrain-lzy (approved):
LGTM

PR #170, opened 2025-05-30
modify cost_model implementation and add debug timer in planner

mikethegoblin:
- added functions and data structures for cost model to collect c2c_planner info from all ranks
- added debug timer to measure the duration of each execution stage of c2c algo


flagcx/core/c2c_algo.cc
```diff
@@ -1507,17 +1590,42 @@ flagcxResult_t flagcxC2cPlanner::execute(const void *sendbuff, void *recvbuff,
     // deviceAdaptor->streamSynchronize(stream);
 
     // execute heteroFuncs
+    if (debugMode_) {
```
aoyulong: It’s preferable to utilize conditional preprocessing without incorporating debug logic directly into the code.

PR #210, opened 2025-08-04
Fix C2C strategy selection when FLAGCX_C2C_ALGO is unset

MC952-arch:
Previously, the C2C strategy detection logic assumed that the environment variable FLAGCX_C2C_ALGO was always set. If unset, the code failed to perform findStrategy function, leading to a setup error.


mikethegoblin (approved):
LGTM

PR #214, opened 2025-08-07
Add DMA-BUF support

MC952-arch:
This PR introduces support for using DMA-BUF for memory registration in RDMA operations.

### DIFF
--- a/flagcx/core/proxy.cc
+++ b/flagcx/core/proxy.cc
@@ -484,12 +484,11 @@ static flagcxResult_t proxyProgressAsync(flagcxProxyAsyncOp **opHead,
                                          flagcxProxyAsyncOp *op,
                                          int *asyncOpCount) {
   int done = 0;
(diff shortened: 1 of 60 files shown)


flagcx/core/proxy.cc
```diff
@@ -477,6 +485,13 @@ static flagcxResult_t proxyProgressAsync(flagcxProxyAsyncOp **opHead,
                                          int *asyncOpCount) {
   int done = 0;
   // flagcxResult_t res = flagcxInternalError;
+  const char *dmaBufEnable = flagcxGetEnv("FLAGCX_DMABUF_ENABLE");
+  int dmaEnabled = 0; // 默认为关闭
```
mikethegoblin: nit: use bool as type for `dmaEnabled`

MC952-arch: done

liuhe-spec (approved):
LGTM

PR #219, opened 2025-08-15
Add chunksize for better algorithm representation

yuejiewang:
* Add `nchunks` and `chunksize` to the algorithm planner representation, so that the same set of algorithm XML files can be used for various communication sizes.

* `test.py` now contains a more comprehensive example of an 8-rank 2-cluster pipelined AllReduce algorithm, which generates the XML algorithm input files for rank 0~7.


flagcx/core/c2c_algo.cc
```diff
@@ -572,21 +572,34 @@ flagcxC2cRefreshFunc::flagcxC2cRefreshFunc()
 flagcxC2cRefreshFunc::flagcxC2cRefreshFunc(size_t offset, size_t count,
                                            size_t totalCount,
                                            flagcxRedOp_t redOp)
-    : offset_(offset), count_(count), totalCount_(totalCount), redOp_(redOp) {}
+    : offset_(offset), count_(count), totalCount_(totalCount), redOp_(redOp) {
```
MC952-arch: Use initializer list for bufftype and start

flagcx/core/c2c_algo.cc
```diff
@@ -572,21 +572,34 @@ flagcxC2cRefreshFunc::flagcxC2cRefreshFunc()
...
+      totalCount_(totalCount), redOp_(redOp) {}
 flagcxC2cRefreshFunc::~flagcxC2cRefreshFunc() {}
 
-flagcxResult_t flagcxC2cRefreshFunc::run(void *buff, flagcxDataType_t datatype,
+flagcxResult_t flagcxC2cRefreshFunc::run(void *recvbuff, void *scratchbuff,
+                                         flagcxDataType_t datatype,
                                          flagcxStream_t stream) {
+  void *refreshbuff = bufftype_ == 1 ? recvbuff : scratchbuff;
+  refreshbuff = static_cast<void *>(static_cast<char *>(refreshbuff) +
+                                    start_ * getFlagcxDataTypeSize(datatype));
```
MC952-arch, line 591: what's the difference between start_ and offset_?

flagcx/core/c2c_algo.cc
```diff
@@ -982,15 +1001,26 @@ flagcxResult_t flagcxC2cPlanner::importXml(const char *path) {
...
   fgets(line, sizeof(line), file);
   nSeqPostSteps_ = readIntTag(line, "nSeqPostSteps");
+  TRACE_CALL("flagcxC2cPlanner import from xml: (nSeqPreSteps, nPipePreSteps, "
+             "nSeqInterSteps, "
+             "nPipePostSteps, nSeqPostSteps) = (%d, %d, %d, %d, %d)",
+             nSeqPreSteps_, nPipePreSteps_, nSeqInterSteps_, nPipePostSteps_,
+             nSeqPostSteps_);
 
   // load refreshFunc
+  int buffType;
```
MC952-arch, line 1010: It’s better to define a specific type, such as flagcxC2cBuffType_t, rather than using a plain int

PR #224, opened 2025-08-22
Fix launch device/host func bug

MC952-arch:


flagcx/core/launch_kernel.h
```diff
@@ -23,47 +23,50 @@ typedef void (*flagcxLaunchFunc_t)(flagcxStream_t, void *);
...
 #endif
 
-struct hostLaunchArgs {
-  volatile bool stopLaunch;
-  volatile bool retLaunch;
+struct flagcxFuncArgs {
+  flagcxStream_t stream;
+  flagcxEvent_t event;
+  bool *recorded;
+  void *value;
```
mikethegoblin, line 37: nit: add a comment explaining the meaning of `value` for better readability. Something like: `value` can be a pointer to hlArgs or dlArgs

mikethegoblin (approved):
LGTM

PR #225, opened 2025-08-23
Add ucx for net_plugin

chencjcj:
[ucx-test.txt](https://github.com/user-attachments/files/21964571/ucx-test.txt)


flagcx/core/net.h
```diff
@@ -17,6 +17,7 @@ typedef char flagcxNetHandle_t[FLAGCX_NET_HANDLE_MAXSIZE];
 #define REGMRBUFFERSIZE (64ULL * 1024 * 1024)
 #define CHUNKSIZE (4ULL * 1024 * 1024)
 #define MAXSTEPS (REGMRBUFFERSIZE / CHUNKSIZE)
+#define FLAGCX_MAX_NET_SIZE_BYTES (1*1024*1024*1024*1024L)
```
MC952-arch, line 21: add a note describing this env

flagcx/core/net_plugin/Makefile
```diff
@@ -0,0 +1,81 @@
+# Makefile for net_ucx_uct plugin
```
MC952-arch: better compile it into libflagcx.so directly rather than using it as a net plugin
flagcx/core/net_ucx_uct.cc

flagcx/service/utils.cc
```diff
@@ -401,3 +401,15 @@ void *flagcxOpenLib(const char *path, int flags,
   }
   return handle;
 }
+
+
+const char *get_plugin_lib_path()
```
MC952-arch, line 406: no need for introducing this func

flagcx/service/check.h
```diff
@@ -50,6 +50,15 @@
     }                                                                          \
   } while (0);
 
+#define PTHREADCHECKGOTO(statement, name, RES, label) do { \
```
MC952-arch, line 53: no need for introducing this macro

flagcx/core/p2p_plugin.h
```diff
@@ -0,0 +1,171 @@
...
+   uint64_t pad[2];
+   struct flagcxIbGidInfo gidInfo;
+ } flagcxIbNetCommDevBase;
+ 
+ enum flagcxIbProvider {
+   IB_PROVIDER_NONE = 0,
+   IB_PROVIDER_MLX5 = 1,
+   IB_PROVIDER_MAX = 2,
+ };
+ typedef struct flagcxIbDev {
```
MC952-arch, line 103: these structs are already defined in net_ib.cc

PR #227, opened 2025-08-27
bug fix: single rank per cluster allgather

yuejiewang:
Fix buffer types for single rank per cluster AllGather operations.


MC952-arch (approved):
LGTM

PR #236, opened 2025-09-05
rename all ibv wrapper function to avoid potential symbol conflict

mikethegoblin:
This PR renames all ibv wrapper function to avoid potential symbol conflict with other linked library. This is a fix for a recent issue where the signature of flagcx's ibv wrapper functions coincided with those declared in another library that FlagCX is linked against, and FlagCX's definitions of these wrapper function overrided the definitions of those in the other library.


yzhang35 (approved):
LGTM

PR #268, opened 2025-10-13
Fix LLaMA3 training resolve device string maca error

whollo:
Fix: resolve maca device string error during LLaMA3 training


yuejiewang (approved):
LGTM

PR #273, opened 2025-10-20
Fix the recv_buffer issue encountered during LLaMA3 training

whollo:
Fix the problem with recv_buffer encountered in the process of LLaMA3 training

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,12 +50,15 @@ jobs:
           apt update -y
           apt-get install clang-format -y
           git fetch --all
-          from_ref=${GITHUB_HEAD_REF}
-          to_ref=${GITHUB_BASE_REF}
-
-          echo "From branch: $from_ref, To branch: $to_ref"
-
-          pre-commit run --from-ref "origin/$from_ref" --to-ref "origin/$to_ref"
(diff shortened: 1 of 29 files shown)


flagcx/core/init.cc
```diff
@@ -216,7 +216,7 @@ static flagcxResult_t flagcxCommInitRankFunc(struct flagcxAsyncJob *job_) {
   }
   FLAGCXCHECK(flagcxNetInit(comm));
   INFO(FLAGCX_INIT, "Using network %s", comm->netAdaptor->name);
-  if (env && strcmp(env, "TRUE") == 0) {
+  if (env && (strcmp(env, "TRUE") == 0 || strcmp(env, "True") == 0)) {
```
MC952-arch, line 219: This env is useless and we can remove it.

flagcx/core/cost_model.cc
```diff
@@ -10,7 +10,8 @@ flagcxResult_t flagcxAlgoTimeEstimator::getAlgoTime(float *time) {
   const char *interServerTopoFile =
       flagcxGetEnv("FLAGCX_INTERSERVER_ROUTE_FILE");
   if (enableTopoDetect && interServerTopoFile &&
-      strcmp(enableTopoDetect, "TRUE") == 0) {
+      (strcmp(enableTopoDetect, "TRUE") == 0 ||
```
MC952-arch, line 13: Try strcasecmp instead

PR #283, opened 2025-10-27
Sync profiling data across ranks

mikethegoblin:
This PR adds allgather among ranks to synchronize the profiling data recorded on each rank. This prevents the tuner to select different communicator as best comm on different ranks due to inconsistencies among profiling data
Currently, the autotuner selects the best communicator based on the profiled latency of specific communication operation; the communicator with the lowest latency will be selected. Since the latency recorded on each rank may differ from one another, we need to take the average of the latency on each rank to make sure each rank selects the same communicator to perform the communication operation.

snapshot from test:
image
image

### DIFF
--- a/flagcx/core/flagcx_tuner.cc
+++ b/flagcx/core/flagcx_tuner.cc
@@ -282,8 +282,6 @@ static flagcxResult_t findBestComm(struct flagcxTunerContext *ctx,
       continue;
     }
 
-    INFO(FLAGCX_TUNING, "before allgather, duration for rank %d is %.3fms",
-         internalTuner.rank, duration);
     memcpy(internalTuner.profilingResults + internalTuner.rank, &duration,
            sizeof(float));
     // get average duration across all ranks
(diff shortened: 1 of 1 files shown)


flagcx/core/flagcx_tuner.cc
```diff
@@ -271,12 +271,33 @@ static flagcxResult_t findBestComm(struct flagcxTunerContext *ctx,
...
       // no profiling data for this communicator and collective category
       WARN("No profiling data for (commId=%d,coll=%d,size=%zu,seq=%u).", idx,
            cat.collType, cat.nBytes, seqId);
       continue;
     }
+
+    INFO(FLAGCX_TUNING, "before allgather, duration for rank %d is %.3fms",
+         internalTuner.rank, duration);
+    float *durationResults = NULL;
+    FLAGCXCHECK(flagcxCalloc(&durationResults, internalTuner.nranks));
```
MC952-arch: pre-allocate this buffer in flagcxTunerInit and de-allocate it in flagcxTunerDestroy

flagcx/core/flagcx_tuner.h
```diff
@@ -7,6 +7,10 @@ struct flagcxTuner {
   // Name of the tuner
   const char *name;
 
+  void *commState;
```
MC952-arch: void *bootstrap

flagcx/core/flagcx_tuner.cc
mikethegoblin: done

flagcx/core/flagcx_tuner.h
mikethegoblin: done

PR #302, opened 2025-11-17
fix vars initialization issue in FlagCX tuner

mikethegoblin:
### PR Category

PAL
### PR Types

Bug Fixes
### PR Description

This PR fixes the issue with empty `vars` vector in flagcx tuner. In the current implementation `vars` is initialized via copy by value; for example, `std::vector<EnvVar> vars = xcclTunerVars;`. However, since `xcclTunerVars` is defined in another translation unit, when we do `std::vector<EnvVar> vars = xcclTunerVars;`, `xcclTunerVars` might not be fully initialized yet, causing `vars` to be an empty vector.
The new implementation uses reference to make sure that `vars` has the same content as the underlying tunerVars at runtime.

### DIFF
--- a/flagcx/adaptor/tuner/tuner.cc
+++ b/flagcx/adaptor/tuner/tuner.cc
@@ -1,11 +1,12 @@
 #include "tuner/tuner_util.h"
 
 #ifdef USE_NVIDIA_ADAPTOR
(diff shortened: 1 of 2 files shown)


MC952-arch (approved):
LGTM

PR #309, opened 2025-11-25
[UIL]Enable flagcxBackend to receive custom backend options

mikethegoblin:
### PR Category

UIL
### PR Types

New Features
### PR Description

This PR enables the flagcxBackend class to receive extra options passed during process group creation. Specifically, the Options class is introduced as a sub-class of flagcxBackend. The Options class currently has one data field called `enableTuner`; this tells flagcx that the current process group can be tuned through auto-tuner. The Options class is pybinded so that we can create the options using python code. A test is done to make sure the options is successfully passed down from python API to C++ API in flagcxBackend.cpp. See below:
image

### DIFF
--- a/plugin/torch/flagcx/src/backend_flagcx.cpp
+++ b/plugin/torch/flagcx/src/backend_flagcx.cpp
@@ -537,15 +537,8 @@ flagcxBackend::allgather(std::vector<std::vector<at::Tensor>> &outputTensors,
 
 #if defined(USE_NVIDIA_ADAPTOR) || defined(USE_METAX_ADAPTOR)
     if (options_->enableTuner && !recordingEnded) {
-      checkRecordingEnded();
-      struct TuneObjectKey tuneObjectKey = {
-          commOpToString(flagcxCommOpAllGather),
-          getDataSize(flagcxDataType, inputTensor.numel())};
-      if (tuneObjectSet_.find(tuneObjectKey) == tuneObjectSet_.end()) {
-        // write this to file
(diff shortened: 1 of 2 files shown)


plugin/torch/example/example.py
```diff
@@ -320,6 +329,7 @@ def test_alltoall():
         print(f"rank {MY_RANK} after all_to_all_single with FLAGCX_GROUP1 (with splits): x = {x}, y = {y}")
 
 def test_all():
+    test_backend_options()
```
MC952-arch: move this part into init_pg?

plugin/torch/example/example.py
```diff
@@ -346,8 +355,10 @@ def test_all():
 if __name__ == "__main__":
     init_pg()
 
-    args = get_args()
+    # args = get_args()
```
MC952-arch: args = get_args()
dict_op_to_test.get(args.op, test_all)()

plugin/torch/flagcx/src/backend_flagcx.cpp
```diff
@@ -1005,6 +1328,20 @@ c10::intrusive_ptr<Work> flagcxBackend::recv(std::vector<at::Tensor> &tensors,
   initComm(tensor.device());
   syncStream(tensor.device());
 
+#if defined(USE_NVIDIA_ADAPTOR) || defined(USE_METAX_ADAPTOR)
```
MC952-arch, line 1217: Encapsulate this section into a function for better modularity

PR #311, opened 2025-11-27
Fix a bug

MC952-arch:
### PR Category
CICD

### PR Types
Bug Fixes & Test Case

### PR Description
This PR fixes a critical bug in the flagcxGroupEnd function and expands test coverage by adding uniRunner mode tests alongside existing homoRunner mode tests.

### DIFF
--- a/flagcx/flagcx.cc
+++ b/flagcx/flagcx.cc
@@ -964,7 +964,7 @@ flagcxResult_t flagcxGroupStart(flagcxComm_t comm) {
 
 flagcxResult_t flagcxGroupEnd(flagcxComm_t comm) {
   if (useHeteroComm()) {
-    FLAGCXCHECK(flagcxRunners[flagcxUniRunner]->groupStart());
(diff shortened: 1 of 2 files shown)


flagcx/flagcx.cc
```diff
@@ -964,7 +964,7 @@ flagcxResult_t flagcxGroupStart(flagcxComm_t comm) {
 
 flagcxResult_t flagcxGroupEnd(flagcxComm_t comm) {
   if (useHeteroComm()) {
-    FLAGCXCHECK(flagcxRunners[flagcxUniRunner]->groupStart());
+    FLAGCXCHECK(flagcxRunners[flagcxUniRunner]->groupEnd());
```
Caozhou1995, line 967: Why didn’t the check function block it? Have you considered enriching the check function?

MC952-arch: FLAGCXCHECK is a debugging helper function, similar to NCCLCHECK.

PR #328, opened 2025-12-19
[CRL] Add one-sided operation support in flagcx with registration and commu…

liuhe-spec:
…nication enhancements

- Implemented `flagcxOneSideRegister` for one-sided memory registration.
- Introduced `flagcxHeteroPut` and `flagcxHeteroPutSignal` for handling one-sided put operations.
- Enhanced `flagcxProxyKernelService` to support new device primitives for one-sided operations.
- Added global variable for managing one-sided operation handles.

### DIFF
--- a/flagcx/core/flagcx_hetero.cc
+++ b/flagcx/core/flagcx_hetero.cc
@@ -87,7 +87,7 @@ flagcxResult_t flagcxHeteroPut(flagcxHeteroComm_t comm, int peer,
 
     uint64_t srcOff = srcOffset;
     uint64_t dstOff = dstOffset;
-    void **gHandles = (void **)&globalOneSideHandles;
+    void **gHandles = (void **)globalOneSideHandles;
     void *request = NULL;
     FLAGCXCHECK(comm->netAdaptor->put(sendComm, srcOff, dstOff, size, srcRank,
                                       dstRank, gHandles, &request));
@@ -118,7 +118,7 @@ flagcxResult_t flagcxHeteroPutSignal(flagcxHeteroComm_t comm, int peer,
     int dstRank = peer;
 
     uint64_t dstOff = dstOffset;
-    void **gHandles = (void **)&globalOneSideHandles;
+    void **gHandles = (void **)globalOneSideHandles;
     void *request = NULL;
     FLAGCXCHECK(comm->netAdaptor->putSignal(sendComm, dstOff, srcRank, dstRank,
                                             gHandles, &request));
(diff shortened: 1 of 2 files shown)


flagcx/adaptor/include/adaptor.h
```diff
@@ -282,7 +282,7 @@ struct flagcxNetAdaptor {
   flagcxResult_t (*put)(void *sendComm, uint64_t srcOff, uint64_t dstOff,
                         size_t size, int srcRank, int dstRank, void **gHandles,
                         void **request);
-  flagcxResult_t (*putSignal)(void *sendComm, uint64_t dstOff, int tag,
+  flagcxResult_t (*putSignal)(void *sendComm, uint64_t dstOff,
```
tengqm, line 285: Wondering if this is a change related to the public interface?
Please be cautious when introduce changes like this ... we are supposed to strive for interface stability whenever possible.

yanminjia: It's not the public interface called by up-layer applications that are changed. Basically, add new interfaces to NET plugin to support one-sided communication interfaces initiated from CPU side.

tengqm: Thanks for the clarification.
LGTM.

flagcx/kernels/device_onesided_demo.cu
```diff
@@ -0,0 +1,59 @@
...
+      } else {
+        __nanosleep(1024);
+      }
+      iter++;
+    }
+    flagcxDeviceTerm(fifoBuffer);
+    flagcxDeviceWait(fifoBuffer);
+  }
+}
+
```
tengqm, line 48: The following code are not supposed to live in this same file.
Please move them to somewhere else.

flagcx/kernels/device_onesided_demo.cu
```diff
@@ -0,0 +1,59 @@
+#include "comm.h"
+#include "flagcx.h"
+#include "flagcx_kernel.h"
+#include "global_comm.h"
+
+FLAGCX_GLOBAL_DECORATOR void flagcxOnesidedSendKernel(const void *srcbuff,
```
tengqm, line 17: OneSided -> Unidirectional?

flagcx/adaptor/net/ibrc_adaptor.cc
```diff
@@ -2467,7 +2467,7 @@ flagcxResult_t flagcxIbPut(void *sendComm, uint64_t srcOff, uint64_t dstOff,
   return flagcxSuccess;
 }
 
-flagcxResult_t flagcxIbPutSignal(void *sendComm, uint64_t dstOff, int tag,
+flagcxResult_t flagcxIbPutSignal(void *sendComm, uint64_t dstOff,
                                  int srcRank, int dstRank, void **gHandles,
```
MC952-arch, line 2471: srcRank is not used in flagcxIbPutSignal

PR #333, opened 2025-12-23
Fix license disclaims

tengqm:
### PR Category

Documentation

### PR Types

Docs

### PR Description

This PR fixes some of the license ambiguity problems found in the project.


plugin/torch/flagcx/include/event_flagcx.hpp
```diff
@@ -1,6 +1,7 @@
 /*************************************************************************
- * Copyright (c) 2025 by MetaX Integrated Circuits (Shanghai) Co., Ltd. All
- *Rights Reserved. Copyright (c) 2025 by DU. All Rights Reserved.
+ * Copyright (c) 2025 by MetaX Integrated Circuits (Shanghai) Co., Ltd.
```
mikethegoblin, line 2: Should we keep vendor license info in torch plugin files? @MC952-arch

tengqm: It is in general okay to leave some copyright info in the source code, though not encouraged.
We'd adovcate for all contributors to refrain from adding copyright text into the source code.

However, copyright is not the same thing as license. The whole project (FlagCX) is licensed under Apache License version 2.0, no matter how the contributors annotate their contributions with copyrights.

We are about to remove these copyright text eventually, but that may happen only after some conversations with the vendors.

mikethegoblin: got it

mikethegoblin (approved):
LGTM

PR #344, opened 2026-01-07
Fix flagcx backend compilation issue for torch version < 2.5.0

mikethegoblin:
### PR Category

UIL
### PR Types

Bug Fixes
### PR Description

This PR fixes flagcx backend compilation issue on Nvidia and MetaX machines when torch version < 2.5.0. When torch version is below 2.5.0, the torch `Backend` class does not have child class `Options`, so backend options should not be enabled in flagcx backend.

### DIFF
--- a/plugin/torch/flagcx/include/backend_flagcx.hpp
+++ b/plugin/torch/flagcx/include/backend_flagcx.hpp
@@ -78,8 +78,8 @@ class flagcxBackend : public Backend {
 public:
 // TODO: check with all vendors to make sure their torch implementation support
 // backend options
-#if defined(USE_NVIDIA_ADAPTOR) ||                                             \
-    defined(USE_METAX_ADAPTOR) && defined(TORCH_VER_GE_250)
+#if (defined(USE_NVIDIA_ADAPTOR) || defined(USE_METAX_ADAPTOR)) &&             \
+    defined(TORCH_VER_GE_250)
   struct TuneObjectKey {
     std::string commOp;
     size_t nBytes;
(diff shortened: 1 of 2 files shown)


plugin/torch/setup.py
```diff
@@ -4,8 +4,10 @@
 # Disable auto load flagcx when setup
 os.environ["TORCH_DEVICE_BACKEND_AUTOLOAD"] = "0"
 from setuptools import setup, find_packages
+from packaging.version import Version, parse as vparse
 
 adaptor_flag = "-DUSE_NVIDIA_ADAPTOR"
+torch_flag = "-DTORCH_VER_LT_250"
```
heavyrain-lzy, line 10: LGTM

plugin/torch/flagcx/include/backend_flagcx.hpp
```diff
@@ -78,7 +78,8 @@ class flagcxBackend : public Backend {
 public:
 // TODO: check with all vendors to make sure their torch implementation support
 // backend options
-#if defined(USE_NVIDIA_ADAPTOR) || defined(USE_METAX_ADAPTOR)
+#if defined(USE_NVIDIA_ADAPTOR) ||                                             \
+    defined(USE_METAX_ADAPTOR) && defined(TORCH_VER_GE_250)
```
MC952-arch: if (defined(USE_NVIDIA_ADAPTOR) || defined(USE_METAX_ADAPTOR)) && defined(TORCH_VER_GE_250)?

mikethegoblin: fixed

PR #356, opened 2026-01-20
try to use a single container file to build different xPU images

SamYuan1990:
### PR Category

CICD

### PR Types

New Features

### PR Description

try to use a single container file to build different xPU images


MC952-arch (approved):
LGTM

PR #363, opened 2026-01-26
add support for ascend

SamYuan1990:
### PR Category

CICD

### PR Types

New Features

### PR Description

Add smoke test for ascend

### DIFF
--- a/.github/workflows/trigger.yml
+++ b/.github/workflows/trigger.yml
@@ -34,7 +34,6 @@ jobs:
               BASE_IMAGE_VERSION=latest
               VENDOR=USE_NVIDIA
           
-          # for now metax base image can't official access without a token.
           # use h20 instance just for build image.
           - job_name: buildcx-metax-image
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger.yml
tengqm, line 30: I believe this is a typo. 'runson' -> 'runs-on'.

SamYuan1990: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#inputs-context
it used for input parameter, the runs-on you are looking for is [here](https://github.com/flagos-ai/build-infra/blob/main/.github/workflows/imagebuild.yml#L51)

ref to github example, in the document, it's either `runson` or `runs_on`

tengqm: Ah... I see.

MC952-arch (approved):
LGTM

PR #364, opened 2026-01-27
Feature/tx81

XingZibo-666:
### PR Category

PAL
### PR Types

New Features
### PR Description

1. TsmicroAdaptor dma support
2. Tsingmicro torch plugin support

### DIFF
--- a/plugin/torch/example/example.py
+++ b/plugin/torch/example/example.py
@@ -1,7 +1,6 @@
 import os
 
 import torch
-backend = "flagcx"
 try:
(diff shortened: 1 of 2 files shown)


plugin/torch/example/example.py
```diff
@@ -1,6 +1,7 @@
 import os
 
 import torch
+backend = "flagcx"
```
MC952-arch: keep using 'f"cpu:gloo,{dev_name}:flagcx"'

PR #365, opened 2026-01-29
Enable launching multi-node API test from one node

mikethegoblin:
### PR Category

UIL
### PR Types

New fetures & Docs
### PR Description

- added python script to launch torch api test across multiple machines based on hostfile and yaml config
- added python script to stop ongoing torch api tests across multiple machines in case we encounter issues that cause the test processes to hang
- added example hostfile and yaml config file for reference
- use local rank when setting device during cross-node test and when there are fewer than 8 GPUs on each node
- updated torch plugin install method to avoid the warning below when using the old install method: `python setup.py develop --adaptor [xxx]`
17640a48-fc16-4b43-b6c1-bcb1426c09b9

example command to launch multi-node torch API test:
`python run_torch_test.py --hostfile example_hostfile --config example_torch_env.yaml`

### DIFF
--- a/plugin/torch/example/run_torch_test.py
+++ b/plugin/torch/example/run_torch_test.py
@@ -15,7 +15,7 @@ import shlex
 import subprocess
 import sys
 import tempfile
-from typing import Dict, List, Tuple
+from typing import Any, Dict, List
 
 try:
     import yaml
@@ -54,9 +54,19 @@ def parse_hostfile(path: str) -> List[Dict[str, str]]:
     return hosts
 
 
-def load_env_config(path: str) -> Tuple[Dict[str, str], Dict[str, Dict[str, str]], str, str, int, str, str, str]:
+def load_env_config(path: str) -> Dict[str, Any]:
     """
-    Reads from yaml config file designated by the user, see example_torch_env.yaml for expected format
+    Reads from yaml config file designated by the user, see example_torch_env.yaml for expected format.
+
+    Returns a dict with keys:
+        - common_env: Dict[str, str]
+        - device_specific_env: Dict[str, Dict[str, str]]
+        - before_start: str
+        - test_dir: str
+        - master_port: int
+        - master_addr: str or None
+        - testfile: str
+        - log_dir: str
     """
     with open(path, "r", encoding="utf-8") as f:
         data = yaml.safe_load(f) or {}
@@ -82,11 +92,13 @@ def load_env_config(path: str) -> Tuple[Dict[str, str], Dict[str, Dict[str, str]
         raise ConfigError("'testfile' must be provided in the config")
     testfile_abs = os.path.abspath(os.path.expanduser(testfile))
 
-    master_port = data.get("master_port", 8281)
+    master_port = data.get("master_port", 12345)
     try:
         master_port = int(master_port)
     except Exception as exc:
         raise ConfigError("'master_port' must be an integer") from exc
+    if not (10000 <= master_port <= 19999):
+        raise ConfigError("'master_port' must be in range 10000-19999")
 
     master_addr = data.get("master_addr", None)
 
@@ -97,7 +109,17 @@ def load_env_config(path: str) -> Tuple[Dict[str, str], Dict[str, Dict[str, str]
     if not isinstance(device_specific, dict):
         raise ConfigError("envs.device_type_specific must be a mapping")
     common = {k: v for k, v in envs.items() if k != "device_type_specific"}
-    return common, device_specific, before_start, test_dir_abs, master_port, master_addr, testfile_abs, log_dir_abs
+
+    return {
+        "common_env": common,
+        "device_specific_env": device_specific,
+        "before_start": before_start,
+        "test_dir": test_dir_abs,
+        "master_port": master_port,
+        "master_addr": master_addr,
+        "testfile": testfile_abs,
+        "log_dir": log_dir_abs,
+    }
 
 
 def validate_hosts(hosts: List[Dict[str, str]], device_specific: Dict[str, Dict[str, str]]):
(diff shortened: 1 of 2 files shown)


plugin/torch/example/stop_torch_test.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Stop torchrun processes across multiple machines",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
+    # Stop processes matching a specific script
+    python stop_torch_test.py --hostfile hosts.txt --pattern "example.py"
```
MC952-arch: run_torch_test.py?

plugin/torch/example/multi_node_script/run_torch_test_1.1.1.2.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
MC952-arch, line 1: how about adding a multi-node script?

plugin/torch/example/stop_torch_test.py
mikethegoblin: `stop_torch_test.py` is intended to be used for stopping torch api tests across multiple nodes. There's a corresponding`run_torch_test.py` for launching torch api test across multiple nodes

plugin/torch/example/example.py
```diff
@@ -56,6 +56,7 @@ def init_pg():
     # Get rank and world_size from environment
     MY_RANK = int(os.environ["RANK"])
     WORLD_SIZE = int(os.environ["WORLD_SIZE"])
+    local_rank = int(os.environ["LOCAL_RANK"])
```
tengqm, line 59: maybe rename `local_rank` to LOCAL_RANK for consistency?

plugin/torch/example/example.py
```diff
@@ -56,6 +56,7 @@ def init_pg():
     # Get rank and world_size from environment
     MY_RANK = int(os.environ["RANK"])
```
tengqm, line 57: Image

plugin/torch/example/example_hostfile
```diff
@@ -0,0 +1,3 @@
+# host slots and device type
+127.0.0.1 slots=8 type=nvidia
```
tengqm, line 2: Why don't we use one of the existing, convenient file format, say INI file?
https://docs.python.org/3/library/configparser.html

plugin/torch/example/run_torch_test.py
```diff
@@ -0,0 +1,215 @@
...
+                    dtype = p.split("=", 1)[1]
+            if slots is None or dtype is None:
+                raise ConfigError(f"Missing slots/type in line: '{line}'")
+            hosts.append({"host": host, "slots": slots, "type": dtype})
+    if not hosts:
+        raise ConfigError("Hostfile is empty")
+    return hosts
+
+
+def load_env_config(path: str) -> Tuple[Dict[str, str], Dict[str, Dict[str, str]], str, str, int, str, str, str]:
```
tengqm: This is ... crazy.
Maybe defining a dict for the return value is much more convenient and less error prone.

plugin/torch/example/example_torch_env.yaml
```diff
@@ -0,0 +1,13 @@
...
+  before_start: source /root/miniconda3/bin/activate flagscale-train
+test_dir: ./
+log_dir: ./torch_logs
+testfile: ./example.py
+master_port: 8281
+master_addr: 127.0.0.1
+envs:
+  FLAGCX_DEBUG: INFO
+  FLAGCX_DEBUG_SUBSYS: ALL
+  device_type_specific:
```
tengqm, line 11: Looks to me we don't need this extra key `device_type_specific`...
We can directly check if `"nvidia" in envs`.

plugin/torch/example/run_torch_test.py
```diff
@@ -0,0 +1,215 @@
...
+    Reads from yaml config file designated by the user, see example_torch_env.yaml for expected format
+    """
+    with open(path, "r", encoding="utf-8") as f:
+        data = yaml.safe_load(f) or {}
+    cmds = data.get("cmds", {})
+    if cmds and not isinstance(cmds, dict):
+        raise ConfigError("cmds must be a mapping")
+    before_start = ""
+    if isinstance(cmds, dict):
+        before_start = cmds.get("before_start", "") or ""
```
tengqm, line 78: why do we add a `or ""` ?

plugin/torch/example/run_torch_test.py
```diff
@@ -0,0 +1,215 @@
...
+    testfile = data.get("testfile", None)
+    if not testfile:
+        raise ConfigError("'testfile' must be provided in the config")
+    testfile_abs = os.path.abspath(os.path.expanduser(testfile))
+
+    master_port = data.get("master_port", 8281)
+    try:
+        master_port = int(master_port)
+    except Exception as exc:
+        raise ConfigError("'master_port' must be an integer") from exc
```
tengqm, line 99: port may need a range (1, 65535)

plugin/torch/example/run_torch_test.py
```diff
@@ -0,0 +1,215 @@
...
+    return """#!/bin/bash
+set -euo pipefail
+{pre_block}{env_exports}
+
+{command}
+""".format(env_exports=env_exports, command=command, pre_block=pre_block)
+
+
+def scp_push(host: str, local_path: str, remote_path: str):
+    subprocess.run(["ssh", host, "mkdir", "-p", os.path.dirname(remote_path)], check=True)
```
tengqm: Maybe using paramiko may help simply our code?

plugin/torch/example/multi_node_script/run_torch_test_1.1.1.2.sh
mikethegoblin: added example multi-node script generated by run_torch_test.py

plugin/torch/example/example_torch_env.yaml
MC952-arch: This key is introduced to handle heterogeneous hardware setups in a structured way, and its value can be parsed directly from the example_hostfile shown above.

plugin/torch/example/example.py
mikethegoblin: local_rank is only used during initialization, it doesn't need to be a global variable

plugin/torch/example/example_hostfile
mikethegoblin: this hostfile format is pretty common for distributed training

plugin/torch/example/run_torch_test.py
mikethegoblin: Thanks for the comment. I changed return type to a dict to make the code tidier.

plugin/torch/example/run_torch_test.py
mikethegoblin: This was a mistake. I removed it.

plugin/torch/example/run_torch_test.py
mikethegoblin: Added port range check to avoid common system occupied port.

plugin/torch/example/run_torch_test.py
mikethegoblin: I wrapped the ssh and scp command into two functions. I think it looks a bit cleaner now.

heavyrain-lzy (approved):
LGTM

PR #370, opened 2026-02-01
[PAL] Support symmetric operation APIs in CCLAdaptor

MC952-arch:
### PR Category
PAL

### PR Types
New Features

### PR Description
This PR adds support for symmetric operation APIs in the CCLAdaptor layer, specifically adding window registration/deregistration functionality for optimized collective operations.

### DIFF
--- a/flagcx/adaptor/ccl/nccl_adaptor.cc
+++ b/flagcx/adaptor/ccl/nccl_adaptor.cc
@@ -30,6 +30,7 @@ flagcxResult_t loadCommOpFuncSymbol(const char *path, const char *name, T *fn) {
   void *sym = dlsym(handle, name);
   if (!sym) {
     fprintf(stderr, "dlsym failed: %s\n", dlerror());
+    dlclose(handle);
     return flagcxSystemError;
   }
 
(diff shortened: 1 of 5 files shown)


flagcx/adaptor/ccl/nccl_adaptor.cc
```diff
@@ -25,22 +117,110 @@ flagcxResult_t ncclAdaptorCommInitRank(flagcxInnerComm_t *comm, int nranks,
                                        flagcxUniqueId_t commId, int rank,
                                        bootstrapState * /*bootstrap*/) {
   if (*comm == NULL) {
-    flagcxCalloc(comm, 1);
+    void *p = malloc(sizeof(struct flagcxInnerComm));
```
mikethegoblin, line 120: why change flagcxCalloc to malloc + memset; I think flagcxCalloc already includes malloc + memset

MC952-arch: It appears that flagcxCalloc only allocates 8 bytes (the size of a pointer), because T is deduced as a pointer type rather than flagcxInnerComm itself.

mikethegoblin: got it

heavyrain-lzy (approved):
LGTM

PR #379, opened 2026-02-09
[CRL] Add custom AllReduce kernels to ncclAdaptor

MC952-arch:
### PR Category
CRL

### PR Types
New Features

### PR Description
Adds built-in (compiled) custom allreduce kernels using NCCL device API and updates the build to generate CUDA -gencode flags based on the detected CUDA version, replacing the prior dynamic symbol-loading approach for allreduce.

### DIFF
--- a/flagcx/kernels/custom_allreduce.cu
+++ b/flagcx/kernels/custom_allreduce.cu
@@ -294,23 +294,28 @@ multimem_st(T* addr, typename packed_t<T, ByteSize>::array_type val) {
   }
 }
 
-// Store to local/shared memory
+// Store to local/shared memory using vectorized store
 template <typename T, int ByteSize = defaultByteSize<T>()>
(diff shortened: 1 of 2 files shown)


flagcx/kernels/custom_allreduce.cu
```diff
@@ -0,0 +1,472 @@
...
+    asm volatile(
+        "multimem.ld_reduce.global.add.f64 %0, [%1];"
+        : "=d"(ret.data[0])
+        : "l"(addr)
+        : "memory");
+  }
+  return ret;
+}
+
+// Min reduction (only supported for 16-bit types: half, bfloat16)
```
ceci3, line 183: Why min and max only support half and bf16 without fp32?

MC952-arch: multimem operation does not support fp32 min/max

heavyrain-lzy (approved):
LGTM

PR #387, opened 2026-02-12
[PAL] Filter out unsupported gencodes to ensure the support of multicast primitive in nvidia_gencode.mk

MC952-arch:
### PR Category
PAL

### PR Types
Bug Fixes


mikethegoblin (approved):
LGTM

PR #389, opened 2026-02-13
[PAL] Extend device primitive to support atomic operations and enable multi-thread producer, single-thread consumer FIFO

MC952-arch:
### PR Category
PAL

### PR Types
Net Features

### DIFF
--- a/flagcx/kernels/flagcx_kernel_device.cu
+++ b/flagcx/kernels/flagcx_kernel_device.cu
@@ -114,14 +114,14 @@ FLAGCX_DEVICE_DECORATOR flagcxResult_t enqueue(void *fifoBuffer, uint64_t addr,
 
   // 5. Build snd value with valid bit set
   uint64_t sndValue =
-      (count & flagcxTriggerMask(flagcxDeviceTriggerBitsCount))
-          << flagcxDeviceTriggerOffCount |
-      (peerRank & flagcxTriggerMask(flagcxDeviceTriggerBitsPeerRank))
-          << flagcxDeviceTriggerOffPeerRank |
-      (datatype & flagcxTriggerMask(flagcxDeviceTriggerBitsDatatype))
-          << flagcxDeviceTriggerOffDatatype |
-      (type & flagcxTriggerMask(flagcxDeviceTriggerBitsPrim))
-          << flagcxDeviceTriggerOffPrim |
+      ((count & flagcxTriggerMask(flagcxDeviceTriggerBitsCount))
+          << flagcxDeviceTriggerOffCount) |
+      ((peerRank & flagcxTriggerMask(flagcxDeviceTriggerBitsPeerRank))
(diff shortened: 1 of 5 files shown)


flagcx/kernels/flagcx_reduce_kernel_device.cu
```diff
@@ -35,26 +35,27 @@ FLAGCX_DEVICE_INLINE_DECORATOR uint64_t flagcxReduceTrigger::getState() {
...
 
-FLAGCX_DEVICE_INLINE_DECORATOR flagcxResult_t dequeue(volatile uint64_t *buffer,
+FLAGCX_DEVICE_INLINE_DECORATOR flagcxResult_t dequeue(uint64_t *buffer,
                                                       int *idx) {
   while (true) {
-    unsigned long long int oldConsumed = *(buffer + 1);
-    unsigned long long int curProduced = *(buffer + 2);
+    uint64_t oldConsumed = flagcxDeviceAtomicLoad(buffer + 1, flagcxDeviceMemoryOrderAcquire);
+    uint64_t curProduced = flagcxDeviceAtomicLoad(buffer + 2, flagcxDeviceMemoryOrderAcquire);
     if (oldConsumed >= curProduced) {
```
yuejiewang, line 49: We don't need atomicLoad here to ensure correctness. Changing these into atomic operations could potentially incur higher costs if there is heavy contention.

MC952-arch: OK

flagcx/kernels/flagcx_kernel_device.cu
```diff
@@ -75,16 +72,16 @@ FLAGCX_DEVICE_DECORATOR flagcxResult_t flagcxDeviceTerm(void *fifoBuffer) {
...
+
+  // Wait until all items are consumed (consumed catches up to produced)
   int iter = 0;
-  while (distance > 0) {
-    spinBackoff(iter);
-    iter++;
-    uint64_t cons = buffer[1];
-    uint64_t prod = buffer[2];
-    distance = prod - cons;
+  while (flagcxDeviceAtomicLoad(&buffer[2], flagcxDeviceMemoryOrderAcquire) >
```
mikethegoblin, line 82: nit: it would be more readable if we name the indices for consumer, producer, and capacity, something like:
```
enum FifoIndex {
FIFO_CAPACITY_IDX = 0,
FIFO_CONSUMER_IDX = 1,
FIFO_PRODUCER_IDX = 2
}
```
Then, instead of using 0, 1, 2 to access the fifo buffer, we can do `buffer[FIFO_CONSUMER_IDX]`, or `buffer[FIFO_PRODUCER_IDX]`.

MC952-arch: A separate PR will be submitted to fix this.

PR #397, opened 2026-02-27
[CRL] Implement unirunner ring reducescatter

yuejiewang:
### PR Category

CRL

### PR Types

New Features

### PR Description

Implement uniRunner ring-sliced reducescatter operation.

### DIFF
--- a/flagcx/runner/uni_runner_impl.cc
+++ b/flagcx/runner/uni_runner_impl.cc
@@ -95,13 +95,6 @@ flagcxResult_t initUniRunnerStateLocRed(flagcxUniRunnerState *runnerState,
         "rank %d initUniRunnerStateLocRed called, count=%lu, numSlices=%d",
         comm->rank, count, numSlices);
 
-  // Initialize queues
-  flagcxIntruQueueConstruct(&runnerState->p2pReadyQueue);
-  flagcxIntruQueueConstruct(&runnerState->redReadyQueue);
-  flagcxIntruQueueConstruct(&runnerState->p2pInflightQueue);
-  flagcxIntruQueueConstruct(&runnerState->redInflightQueue);
-  runnerState->numPendingNodes = 0;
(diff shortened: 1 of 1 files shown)


flagcx/runner/uni_runner_impl.cc
```diff
@@ -968,6 +968,238 @@ static flagcxResult_t initUniRunnerStateSlicedAR(
...
+  flagcxIntruQueueConstruct(&runnerState->p2pReadyQueue);
+  flagcxIntruQueueConstruct(&runnerState->redReadyQueue);
+  flagcxIntruQueueConstruct(&runnerState->p2pInflightQueue);
+  flagcxIntruQueueConstruct(&runnerState->redInflightQueue);
+  runnerState->numPendingNodes = 0;
+
+  int rank = comm->rank;
+  int nranks = comm->nranks;
+
+  if (nranks < 2) {
```
mikethegoblin: Do we need to handle single rank case?

MC952-arch: It would be better to handle this case at the beginning of the function and return flagcxSuccess directly.

flagcx/runner/uni_runner_impl.cc
```diff
@@ -968,6 +968,238 @@ static flagcxResult_t initUniRunnerStateSlicedAR(
...
+        }
+      }
+    }
+  }
+
+  TRACE(FLAGCX_UNIRUNNER,
+        "DAG scheduler initialized with %d-rank ReduceScatter topology (%d "
+        "slices, %d redSlices)",
+        nranks, numSlices, numRedSlices);
+#ifdef UNIRUNNER_PRINT_DAG
```
MC952-arch: Is it necessary to add this compilation flag?

MC952-arch (approved):
LGTM

PR #410, opened 2026-03-11
[PAL] Adaptor plugin mechanism implementation phase1

mikethegoblin:
### PR Category

PAL
### PR Types

New Features
### PR Description

This PR is phase 1 of our adaptor plugin mechanism implementation. Specifically, this PR makes the following changes:
- separate ccl, device, and net adaptor declaration to 3 separate header files, so that the user can include specific header files for their plugin implementation
- extracted common and helper functions that need to be implemented in later phases


yzhang35 (approved):
LGTM

flagcx/adaptor/adaptor_plugin_init.cc
```diff
@@ -0,0 +1,27 @@
...
+//   net/net_plugin_load.cc
+
+flagcxResult_t flagcxAdaptorPluginInit() {
+  FLAGCXCHECK(flagcxCCLAdaptorPluginLoad());
+  FLAGCXCHECK(flagcxDeviceAdaptorPluginLoad());
+  FLAGCXCHECK(flagcxNetAdaptorPluginLoad());
+  return flagcxSuccess;
+}
+
+flagcxResult_t flagcxAdaptorPluginFinalize() {
```
MC952-arch: flagcxAdaptorPluginInit() is wired into flagcxInit(), but there's no corresponding call to flagcxAdaptorPluginFinalize() during shutdown. Libraries loaded via dlopen in future phases will leak their handles.

flagcx/adaptor/include/flagcx_device_adaptor.h
```diff
@@ -0,0 +1,115 @@
...
+  flagcxResult_t (*launchKernel)(void *func, unsigned int block_x,
+                                 unsigned int block_y, unsigned int block_z,
+                                 unsigned int grid_x, unsigned int grid_y,
+                                 unsigned int grid_z, void **args,
+                                 size_t share_mem, void *stream,
+                                 void *memHandle);
+  flagcxResult_t (*copyArgsInit)(void **args);
+  flagcxResult_t (*copyArgsFree)(void *args);
+  flagcxResult_t (*launchDeviceFunc)(flagcxStream_t stream,
+                                     void (*fn)(flagcxStream_t, void *),
```
MC952-arch: Original in adaptor.h: flagcxResult_t (*launchDeviceFunc)(flagcxStream_t stream, flagcxLaunchFunc_t fn, void *args); Why not keep using flagcxLaunchFunc_t?

flagcx/adaptor/plugin_common.cc
```diff
@@ -0,0 +1,32 @@
...
+    return NULL;
+  }
+  void *handle = dlopen(name, RTLD_NOW | RTLD_LOCAL);
+  if (handle == NULL) {
+    INFO(FLAGCX_INIT, "ADAPTOR/Plugin: dlopen(%s) failed: %s", name, dlerror());
+  }
+  return handle;
+}
+
+void *flagcxAdaptorOpenPluginLib(const char *path) {
```
MC952-arch: flagcxAdaptorOpenPluginLib(path) just calls flagcxAdaptorTryOpenLib(path) with identical arguments and return type. Consider collapsing into one function, or documenting the intended distinction (e.g., TryOpen returns NULL on failure for probing, OpenPlugin will log errors and is the primary entry point).

flagcx/adaptor/adaptor_plugin_init.cc
```diff
@@ -0,0 +1,27 @@
...
+#include "core.h"
+
+// Top-level orchestrator for adaptor plugin loading.
+// Per-type load/unload functions are implemented in:
+//   ccl/ccl_plugin_load.cc
+//   device/device_plugin_load.cc
+//   net/net_plugin_load.cc
+
+flagcxResult_t flagcxAdaptorPluginInit() {
+  FLAGCXCHECK(flagcxCCLAdaptorPluginLoad());
```
MC952-arch: If flagcxCCLAdaptorPluginLoad() fails, flagcxDeviceAdaptorPluginLoad() and flagcxNetAdaptorPluginLoad() are skipped entirely. In Phase 2+ when these do real work, a user who only wants a net plugin would be blocked by an unrelated CCL plugin failure.

flagcx/adaptor/plugin_common.cc
```diff
@@ -0,0 +1,32 @@
...
+  return flagcxAdaptorTryOpenLib(path);
+}
+
+flagcxResult_t flagcxAdaptorClosePluginLib(void *handle) {
+  if (handle != NULL) {
+    dlclose(handle);
```
MC952-arch, line 25: dlclose(handle) can fail (returns non-zero). The return value is silently discarded

mikethegoblin: a failed dlclose is non-recoverable, so we can't retry closing a library. dlclose return values are normally ignored in this case

flagcx/adaptor/adaptor_plugin_init.cc
mikethegoblin: I removed the high-level `flagcxAdaptorPluginInit` and `flagcxAdaptorPluginFinalize` because each adaptor plugin should handle their own lifecycle. We'll add calls to individual init and finalize when we implement the init and finalize functions for each adaptor plugin in later phases

flagcx/adaptor/include/flagcx_device_adaptor.h
mikethegoblin: fixed

flagcx/adaptor/plugin_common.cc
mikethegoblin: collapsed into one function

flagcx/adaptor/adaptor_plugin_init.cc
mikethegoblin: removed high-level init and finalize calls. We'll call inidividual plugin init and finalize functions and do error handling in these individual init and finalize functions

MC952-arch (approved):
LGTM

PR #414, opened 2026-03-16
[PAL] FlagCX adaptor plugin mechanism phase 2: Net Adaptor Plugin

mikethegoblin:
### PR Category

PAL
### PR Types

New Features
### PR Description

This PR implements the net adaptor plugin loading mechanism and modify existing code to enable using loaded plugin during network communication


flagcx/adaptor/net/net_plugin_load.cc
```diff
@@ -3,46 +3,91 @@
...
+    flagcxAdaptorClosePluginLib(netPluginDlHandle);
+    netPluginDlHandle = NULL;
+    return flagcxSuccess;
+  }
+
+  // Validate critical function pointers
+  if (plugin->name == NULL || plugin->init == NULL || plugin->devices == NULL ||
+      plugin->listen == NULL || plugin->connect == NULL ||
+      plugin->accept == NULL || plugin->isend == NULL ||
+      plugin->irecv == NULL) {
```
MC952-arch: Incomplete function pointer validation in plugin loader.
if (plugin->name == NULL || plugin->init == NULL || plugin->devices == NULL ||
      plugin->getProperties == NULL || plugin->listen == NULL ||
      plugin->connect == NULL || plugin->accept == NULL ||
      plugin->closeSend == NULL || plugin->closeRecv == NULL ||
      plugin->closeListen == NULL || plugin->regMr == NULL ||
      plugin->deregMr == NULL || plugin->isend == NULL ||
      plugin->irecv == NULL || plugin->test == NULL) {

flagcx/adaptor/net/net_plugin_load.cc
```diff
@@ -3,46 +3,91 @@
...
+
+  flagcxNetAdaptors[0] = plugin;
+  INFO(FLAGCX_INIT, "ADAPTOR/Plugin: Loaded net adaptor plugin '%s'",
+       plugin->name);
+  return flagcxSuccess;
+}
+
+flagcxResult_t flagcxNetAdaptorPluginUnload() {
+  flagcxNetAdaptors[0] = nullptr;
+  flagcxAdaptorClosePluginLib(netPluginDlHandle);
```
MC952-arch, line 79: flagcxNetStates has been removede?

flagcx/adaptor/net/net_plugin_load.cc
```diff
@@ -3,46 +3,91 @@
...
+    return flagcxSuccess;
+  }
+
+  flagcxNetAdaptors[0] = plugin;
+  INFO(FLAGCX_INIT, "ADAPTOR/Plugin: Loaded net adaptor plugin '%s'",
+       plugin->name);
+  return flagcxSuccess;
+}
+
+flagcxResult_t flagcxNetAdaptorPluginUnload() {
```
MC952-arch: flagcxNetAdaptorPluginLoad and flagcxNetAdaptorPluginUnload are non-static file-scope functions. They are only intended to be called from flagcxNetAdaptorPluginInit/Finalize (which hold netPluginMutex), but being externally visible means they could be called directly without the mutex, creating a thread-safety gap. Either make them static or move the mutex acquisition into them.

adaptor_plugin/net/example/plugin.cc
```diff
@@ -0,0 +1,178 @@
+/*************************************************************************
```
MC952-arch, line 1: The example plugin gets FLAGCX_PTR_HOST/FLAGCX_PTR_CUDA/FLAGCX_PTR_DMABUF via the chain flagcx_net_adaptor.h → flagcx.h, which works because the symlinks point to the real headers. For out-of-tree plugins that copy headers, this implicit dependency is easy to miss.

mikethegoblin: the pointer types are defined in `flagcx_net.h`, which the user is required to import for implementing the plugin, so this is not a problem. The plugin is designed to make the user include the minimum amount of header files from FlagCX (currently `flagcx.h`, `flagcx_net_adaptor.h`, `net_device.h`, `flagcx_net.h`). These 4 header files do not have extra dependencies and are sufficient to provide the users with the necessary data structures and API definitions they need to implement the plugin.

flagcx/adaptor/net/net_plugin_load.cc
mikethegoblin: fixed

flagcx/adaptor/net/net_plugin_load.cc
mikethegoblin: fixed

flagcx/adaptor/net/net_plugin_load.cc
mikethegoblin: fixed

MC952-arch (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #415, opened 2026-03-16
[CRL] Align one-sided RMA with NCCL: chained put+signal

MC952-arch:
### PR Category
CRL

### PR Types
New Features

### PR Description
This PR updates FlagCX’s one-sided RMA path to more closely mirror NCCL GIN semantics by introducing a chained data-write + signal (atomic) operation and splitting data vs. signal memory registrations, while also expanding the device API surface (requirements struct, cooperative groups, properties query).


mikethegoblin (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #428, opened 2026-03-24
[UIL] add auto platform detection for pip install

mikethegoblin:
### PR Category

UIL
### PR Types

New Features
### PR Description

This PR introduces automatic platform detection mechanism during flagcx pip build. This frees the user from manually specifying the platform through env variable when running `pip install`. The mechanism works by checking if the platform-specific command like `nvidia-smi` exists to determine which platform the user is currently operating on. Now, a user can simply run `pip install . --no-build-isolation` to build the vanilla version of flagcx with no advanced feature included. The user can still choose to explicitly specify env variables to enable advanced feature for the build.

### DIFF
--- a/plugin/torch/_build_config.py
+++ b/plugin/torch/_build_config.py
@@ -55,6 +55,7 @@ _PLATFORM_COMMANDS = [
     ("mx-smi", "metax"),
     ("hy-smi", "du"),
     ("xpu-smi", "klx"),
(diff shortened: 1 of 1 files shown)


MC952-arch (approved):
LGTM

heavyrain-lzy (approved):
LGTM

PR #433, opened 2026-03-26
[PAL] Introduce Device Traits abstraction and DeviceAPI for unified vendor/fallback support

MC952-arch:
### PR Category
PAL

### PR Types
New Features


flagcx/adaptor/include/flagcx_ccl_adaptor.h
```diff
@@ -105,6 +107,13 @@ struct flagcxCCLAdaptor_v1 {
   // Group semantics
   flagcxResult_t (*groupStart)();
   flagcxResult_t (*groupEnd)();
+
```
mikethegoblin, line 110: adding new APIs to the v1 version of flagcxCCLAdaptor would break the example ccl plugin implementation. We can either:
1. modify the example plugin implementation to make sure the example CCL plugin can still be built and loaded correctly
2. add flagcxCCLAdaptor_v2 and put the two new APIs in the v2 struct, then implement backward compatibility mechanism in ccl plugin loader to convert v1 struct to v2 struct.  

Approach 1 requires the least change; approach 2 might need some modification to the plugin loading logic

flagcx/adaptor/include/flagcx_ccl_adaptor.h
```diff
@@ -14,6 +14,8 @@ extern "C" {
 // Forward declarations for opaque types not in flagcx.h
 // (flagcxStream_t, flagcxWindow_t are already typedef'd in flagcx.h)
 typedef struct flagcxInnerComm *flagcxInnerComm_t;
+typedef struct flagcxInnerDevComm *flagcxInnerDevComm_t;
+struct flagcxDevCommRequirements;
```
mikethegoblin, line 18: We will need to also include `flagcx_kernel.h` as a header file for the CCL plugins to use flagcxDevCommRequirements properly because plugins are built separately from libflagcx.so and would hence need to see full definition of a struct to use it. We can open another PR to fix the plugin dependency issue

MC952-arch: Can we defer it to V2?

yuejiewang (approved):
LGTM

flagcx/adaptor/include/flagcx_ccl_adaptor.h
mikethegoblin: sure

mikethegoblin (approved):
LGTM

PR #459, opened 2026-04-23
[UIL] Support allgather with different message size in torch flagcx backend

yanminjia:
### PR Category

UIL
### PR Types

New Feature
### PR Description

Support allgather with different message size in torch flagcx backend. In case of MoE training with TP enabled, after dispatch communication (post_dispatch) in the forward pass, the dispatched tokens (different number) should be aggregated in TP communication group via allgather with different size message.


MC952-arch (approved):
LGTM

mikethegoblin (approved):
LGTM

PR #470, opened 2026-05-09
[CRL] Reuse closed connection slots to fix proxy destroy hang

MC952-arch:
### PR Category
CRL

### PR Types
Bug Fixes

### PR Description
This PR updates the proxy service thread’s accept logic to reuse previously closed connection slots (identified by pollfds[].fd == -1), which should prevent hangs during proxy teardown when the stop connection can’t be accepted due to exhausted slot indices.


mikethegoblin (approved):
LGTM

yuejiewang (approved):
LGTM

PR #491, opened 2026-06-04
[CRL] Bootstrap extension

MC952-arch:
### PR Category
CRL

### PR Types
New Features

### PR Description
This PR extends the bootstrap layer to support a new P2P mode (RPC-style listen/connect/accept) alongside the existing collective/ring bootstrap, and wires that into the IB P2P engine so control-plane metadata + descriptor-table exchange can occur over a bootstrap socket rather than over the IB data socket.

### DIFF
--- a/flagcx/core/flagcx_p2p.cc
+++ b/flagcx/core/flagcx_p2p.cc
@@ -12,6 +12,7 @@
 #include "flagcx_p2p.h"
 
 #include "adaptor.h"
+#include "bootstrap.h"
 #include "debug.h"
 #include "flagcx_net.h"
 #include "flagcx_net_adaptor.h"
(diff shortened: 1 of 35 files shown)


flagcx/core/flagcx_p2p.cc
```diff
@@ -1833,30 +1942,55 @@ FlagcxP2pConn *flagcxP2pEngineConnect(FlagcxP2pEngine *engine,
...
+  }
+
+  // Step 2: Exchange IB listen handles over bootstrap
+  char localIbHandle[FLAGCX_NET_HANDLE_MAXSIZE];
+  memcpy(localIbHandle, engine->listeners[netDev].handle,
+         FLAGCX_NET_HANDLE_MAXSIZE);
+
+  char remoteIbHandle[FLAGCX_NET_HANDLE_MAXSIZE];
+  memset(remoteIbHandle, 0, sizeof(remoteIbHandle));
+  if (bootstrapExchange(bsConn, 0, 4, localIbHandle, FLAGCX_NET_HANDLE_MAXSIZE,
```
mikethegoblin, line 1971: `bootstrapExchange` in P2P mode now does send followed by recv on both peers. Since both connector and acceptor call the same exchange path, they can both block in send if the payload
  exceeds socket buffering, and neither side reaches recv to drain the other peer’s data. This is especially risky for descriptor-table exchange.

PR #502, opened 2026-06-15
[PAL] fix p2p socket destruction issue

mikethegoblin:
### PR Category

PAL
### PR Types

Bug Fixes
### PR Description

This PR fixes the socket destruction problem:
#### Root Cause
The P2P engine shutdown path closed and freed bootstrap/IB listen sockets while an accept loop could still be blocked on them. During teardown, accept() could observe a closed or
  invalid listener fd and return EINVAL/EBADF, which surfaced as noisy socket warnings even though data transfer had already completed successfully.

#### Solution
This PR adds an explicit accept-cancellation path for the FlagCX P2P engine. flagcxP2pEngineStopAccept() now signals accept shutdown, wakes bootstrap and IB accept loops via abort-
  enabled listen sockets, and joins the internal RPC accept thread before final resource cleanup. Listener resources are only freed during engine destruction after accept users have
  exited. The socket layer also treats abort-triggered accept failures as expected cancellation, avoiding false warning logs during normal teardown.

### DIFF
--- a/flagcx/core/flagcx_p2p.cc
+++ b/flagcx/core/flagcx_p2p.cc
@@ -44,6 +44,7 @@
 #include <unistd.h>
 
 extern struct flagcxNetAdaptor flagcxNetIbP2p;
+extern flagcxResult_t flagcxNetIbP2pAbortListen(void *listenComm);
 
 extern "C" flagcxResult_t flagcxP2pSliceBatch(void *sendComm, struct ibv_qp *qp,
                                               int count, FlagcxSlice **slices);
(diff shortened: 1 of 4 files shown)


flagcx/core/flagcx_p2p.cc
```diff
@@ -2047,6 +2059,8 @@ FlagcxP2pConn *flagcxP2pEngineAccept(FlagcxP2pEngine *engine, char *ipAddrBuf,
                                      size_t ipAddrBufLen, int *remoteGpuIdx) {
   if (engine == NULL || ipAddrBuf == NULL || remoteGpuIdx == NULL)
     return NULL;
+  if (engine->stopAccept.load(std::memory_order_acquire))
```
MC952-arch, line 2062: either use std::atomic::load/store or __atomic_load/store_n

mikethegoblin: `engine->stopAccept` is of type `std::atomic<bool>`, so the load operation here is using `std::atomic::load`

PR #507, opened 2026-06-29
[PAL] Allow send/recv operation to be invoked individually without group operation in hccl adaptor

mikethegoblin:
### PR Category

PAL
### PR Types

Bug Fixes
### PR Description

This PR fixes the problem where hccl send/recv operations must be called in pairs and within a group.


MC952-arch (approved):
LGTM

PR #510, opened 2026-07-03
[CRL] Lazy initialize P2P engine global vars

mikethegoblin:
### PR Category

CRL
### PR Types

Others
### PR Description

FlagCX P2P engine contains global variables that will be initialized even when we don't use it (e.g. when performing collective operations). This PR implements lazy initialization of P2P engine related global variables, so that when P2P engine is not used, those global variables will not be initialized


MC952-arch (approved):
LGTM

PR #530, opened 2026-08-06
fix(packaging): use system nlohmann-json

shiptux:
## Summary

- make the nlohmann-json include root configurable while preserving the git submodule default for source builds
- build Debian and RPM packages against `/usr/include` and declare the matching distro build dependency
- fix RPM dependency selection: Fedora and RHEL/EPEL use `json-devel`; openEuler uses `nlohmann-json-devel`

## Verification

- parsed `setup.py` with Python AST
- confirmed a FlagCX make dry-run emits `-I/usr/include`
- parsed the RPM spec for Fedora 43, RHEL 9, and openEuler macro sets
- verified package availability: Fedora 43/Rocky 9 provide `json-devel 3.11.3`; openEuler 24.03 provides `nlohmann-json-devel 3.11.2`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,6 +226,8 @@ BUILD_INCDIR := $(BUILDDIR)/include
 PREFIX ?= /usr/local
 DESTDIR  ?= $(PREFIX)/lib
 INC_DESTDIR ?= $(PREFIX)/include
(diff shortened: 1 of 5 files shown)


MC952-arch (approved):
LGTM

PR #552, opened 2026-08-21
[PAL] support latest enflame topsPlatform 4.0 and Host API perf test

fanfanfanWu:
PR Category
[PAL]

PR Types
Fix

PR Description
fix compile error for Enflame topsplatform 4.0
fix Enflame host api test


makefiles/enflame.mk
```diff
@@ -19,4 +19,4 @@ ADAPTOR_FLAG := -DUSE_ENFLAME_ADAPTOR
 
 PLATFORM_KERNEL_DIR  :=
 PLATFORM_KERNEL_SRCS :=
-PLATFORM_EXTRA_SRCS  :=
+PLATFORM_EXTRA_SRCS  := flagcx/adaptor/device_api/default_dev_api_backend.cc
```
MC952-arch, line 22: **Fix location: `flagcx/adaptor/device_api/default_dev_api_backend.cc:682-719`.**

This line makes a shared default-backend cleanup bug reachable on Enflame; the defect is not Enflame-specific. With `FLAGCX_SIGNAL_HOST_ENABLE=1`, the common backend unregisters the local SHM mapping through `peerBarrierShmPtrs`, unregisters it again through `localBarrierShmPtr`, and then passes the mapped host VA in `localBarrierFlags` to `deviceFree`.

Please fix the common teardown so each host registration is released exactly once and the SHM-mapped VA is excluded from the generic device-allocation cleanup. This also protects other adaptors that combine the default backend with real host registration.

flagcx/adaptor/device/tops_adaptor.cc
```diff
@@ -385,10 +385,21 @@ flagcxResult_t topsAdaptorStreamWriteValue64(flagcxStream_t, void *, uint64_t,
   return flagcxNotSupported;
 }
 
-flagcxResult_t topsAdaptorHostRegister(void *, size_t) {
-  return flagcxNotSupported;
+flagcxResult_t topsAdaptorHostRegister(void *ptr, size_t size) {
+  if (ptr == NULL || size == 0) {
+    return flagcxInvalidArgument;
+  }
+  DEVCHECK(topsHostRegister(ptr, size, topsHostRegisterMapped));
```
MC952-arch, line 392: **Fix location: `flagcx/adaptor/device_api/default_dev_api_backend.cc:248-300`.**

Implementing a real `topsHostRegister` exposes a second shared default-backend ownership bug; the TOPS wrapper itself is not where the cleanup should live. If registration or device-pointer lookup fails partway through the peer loop, `fail_peer_shm` calls `hostUnregister` for every mapped peer, including entries that were never successfully registered.

Please track successful registrations in the common backend, counting the current entry immediately after registration succeeds, and unregister only that set while preserving the original initialization error.

MC952-arch (changes requested):
The TOPS 4.0 compatibility changes look reasonable, but enabling real host registration together with the default device-API backend exposes invalid registration cleanup during communicator initialization and destruction. Please correct the host-mapping ownership paths before merging.

MC952-arch (approved):
LGTM

flagcx/adaptor/device/tops_adaptor.cc
MC952-arch: Refer to https://github.com/flagos-ai/FlagCX/pull/553

makefiles/enflame.mk
MC952-arch: Refer to https://github.com/flagos-ai/FlagCX/pull/553

PR #555, opened 2026-08-24
klx Add allreduce alltoall

sunge666-ui:
### PR Category
<CRL (Communication Runtime Layer)>

### PR Types
< [ New Features |>

### PR Description
Integrated xshmem into shmemadaptor on the host side and implemented `allreduce` and `alltoall_oneside` for the device side; verified functionality via `perf_test` with data sizes reaching 1 GB.


flagcx/adaptor/shmem/shmem_adaptor.h
```diff
@@ -25,7 +25,7 @@ struct flagcxShmemAdaptor {
   const char *name;
 
   // Lifecycle (reference-counted)
-  flagcxResult_t (*init)(int rank, int nRanks);
+  flagcxResult_t (*init)(flagcxComm_t comm);
```
MC952-arch: why we have to modify this function signature?

ssuurrffaaccee: xshmem_init function need bkclcontext; flagcxComm holds  bkclcontexnt instance. code: https://github.com/flagos-ai/FlagCX/pull/555/changes#diff-03a9471c3aee61b7fd71ab4dc4951ed310b348345f41b30052862eea228e94a5R37

makefiles/kunlunxin.mk
```diff
@@ -19,4 +23,11 @@ ADAPTOR_FLAG := -DUSE_KUNLUNXIN_ADAPTOR
 
 PLATFORM_KERNEL_DIR  := flagcx/adaptor/kernel/kunlunxin
 PLATFORM_KERNEL_SRCS := $(wildcard $(PLATFORM_KERNEL_DIR)/*.$(DEVICE_FILE_EXTENSION))
-PLATFORM_EXTRA_SRCS  :=
+
+ifeq ($(USE_XSHMEM), 1)
```
MC952-arch: **Fix location: `makefiles/kunlunxin.mk:27-33`.**

The repository-wide selector is `USE_SHMEM`: the root Makefile defines it and `test/make.inc` uses it to enable `FLAGCX_COMM_TRAITS_SHMEM`. `USE_XSHMEM` is not defined or forwarded elsewhere, so the normal KunlunXin SHMEM build falls through to `default_dev_api_backend.cc` and omits both new XSHMEM sources while tests compile the XSHMEM path. Use `USE_SHMEM` here, or consistently introduce and propagate a separate documented selector.

test/kernel/klx/device_api.xpu
```diff
@@ -0,0 +1,588 @@
...
+
+    long nt0 = t0 + tile_elems;
+    if (nt0 < num_elems) {
+      int ntlen = (int)((num_elems - nt0 < tile_elems) ? (num_elems - nt0)
+                                                       : tile_elems);
+      DeviceAPI::Window nextData;
+      nextData.symBase = io_data + nt0;
+      nextData.rawPtr = io_data + nt0;
+      nextData.allocSize = (size_t)ntlen * sizeof(float);
+      issueSegPuts(net, world, ((seq + 1) & 1) ? pongWin : pingWin, nextData,
```
MC952-arch: **Fix locations: `test/kernel/klx/device_api.xpu:264-291`, `test/kernel/klx/device_api.xpu:131-153`.**

The arrival signal proves that peers finished putting tile `t`; it does not prove they finished reducing the data that this rank put into their tile-`t` slot. A fast rank can reach tile `t+1` and pre-put tile `t+2` into the same ping/pong slot while a slower peer is still in `reduceSeg` for tile `t`. Add a consumption acknowledgement/barrier before a slot is reused, and test at least three tiles with skewed ranks.

test/kernel/klx/device_api.xpu
```diff
@@ -0,0 +1,588 @@
...
+    if (nt0 < count) {
+      int ntlen = min(tile_elems, count - nt0);
+      int ns, ne;
+      tile_range(sub_clusterid, sub_clusternum, stride, ntlen, &ns, &ne);
+      __global_ptr__ float *next_stage =
+          ((seq + 1) & 1) ? stage_pong : stage_ping;
+      dstWin.symBase = next_stage;
+      dstWin.rawPtr = next_stage;
+      if (ne > ns) {
+        net.put(world, dst_pe, dstWin,
```
MC952-arch: **Fix location: `test/kernel/klx/device_api.xpu:425-498`.**

This AlltoAll pipeline has the same missing consumption acknowledgement: the signal only covers put completion. Two iterations later, a fast source can put into the same stage slot before the destination has finished copying the older tile out of it. Synchronize consumption before slot reuse rather than relying only on ping/pong alternation.

makefiles/kunlunxin.mk
```diff
@@ -4,9 +4,13 @@
...
-DEVICE_LINK  := -lxpurt -lcudart
-DEVICE_PLATFORM :=
-DEVICE_COMPILER :=
+DEVICE_LINK  := -lxpurt -lcudart -lxpucuda -lxpuml
+# DEVICE_PLATFORM selects the test kernel subdir (test/kernel/<lowercase>): klx.
+DEVICE_PLATFORM := KLX
+# XPU (XTDK) clang toolchain — only used to compile test .xpu kernels
+# (COMPILE_KERNEL / test/kernel/klx). The main libflagcx.so build uses g++.
+XTDK_HOME ?= /workspace/my_flagcx/xtdk-llvm15-ubuntu2004_x86_64
+DEVICE_COMPILER := $(XTDK_HOME)/bin/clang++
```
MC952-arch, line 17: **Fix locations: `makefiles/kunlunxin.mk:10-15`, `Makefile:357-364`.**

The root kernel rule relies on `DEVICE_COMPILE_FLAG` to provide `-c`, but it is empty here. With `DEVICE_COMPILER=clang++`, `COMPILE_KERNEL=1` therefore attempts to link each `.xpu` source into an `.o`, and it also omits the `--xpu-arch=xpu3` flag that the new KLX test rule requires. Configure both compile and supported device-link flags (the root also invokes `-dlink`) or separate the test-only toolchain path.

test/kernel/klx/Makefile
```diff
@@ -0,0 +1,47 @@
...
+all: $(OBJDIR)/device_api.o
+
+device_api.o: $(OBJDIR)/device_api.o
+
+$(OBJDIR)/device_api.o: device_api.xpu
+	@mkdir -p $(OBJDIR)
+	@echo "Compiling device_api.xpu (KLX / native xshmem)"
+	@$(XPU_CLANG) $(XPU_CFLAGS) $(EXTRA_XPU_FLAGS) $(ALL_INCLUDES) $(XPU_DEVLINK) -c -o $@ $<
+
+# The native xshmem path performs its own device-link inline (above), so the
```
MC952-arch: **Fix locations: `test/kernel/klx/Makefile:31-43`, `test/kernel/klx/device_api.xpu:29-37`.**

These no-op rules return success without creating the requested files, but `test/unittest/device_api` later links `device_ir.o`/`dlink_ir.o`. Separately, this platform's `device_api.o` defines only two entry points while the perf wildcard includes the two-sided benchmark and the Device API unit tests reference many more. Normal KLX test builds therefore fail at link time. Produce the required artifacts/symbols or filter unsupported targets explicitly.

flagcx/adaptor/shmem/xshmem_adaptor.cc
```diff
@@ -0,0 +1,193 @@
...
+static flagcxResult_t xshmemAdaptorInit(flagcxComm_t comm) {
+#ifdef USE_KUNLUNXIN_ADAPTOR
+  if (comm == nullptr || comm->homoComm == nullptr)
+    return flagcxInternalError;
+
+  // xccl xshmem initializes against the BKCL context created by
+  // bkcl_init_rank (stored in flagcxInnerComm::base). There is no no-arg
+  // xshmem_init(), so this is the only way to obtain the context.
+  BKCLContext_t ctx = comm->homoComm->base;
+  if (xshmem_init(ctx) != 0)
```
MC952-arch: **Fix location: `flagcx/adaptor/shmem/xshmem_adaptor.cc:25-52`.**

The local reference count does not control XSHMEM: every call invokes `xshmem_init(ctx)`, while `finalize` only decrements the integer and never releases/resets the runtime. A rank/size mismatch after successful init also returns without cleanup. This is exercised by `test_multi_fifo`, which repeatedly creates and destroys device communicators. Make first/last-reference behavior real (or deliberately use one process-lifetime init), handle mismatch cleanup, and synchronize the state.

test/kernel/klx/device_api.xpu
```diff
@@ -0,0 +1,588 @@
...
+
+namespace {
+struct XshmemA2AScratch {
+  float *stage = nullptr;
+  uint64_t *signal = nullptr;
+  long stageElems = 0;
+  int clusternum = 0;
+  int npes = 0;
+};
+static XshmemA2AScratch g_a2a;
```
MC952-arch: **Fix locations: `test/kernel/klx/device_api.xpu:58-99`, `test/kernel/klx/device_api.xpu:502-585`.**

These mutable scratch allocations and sequence counters are shared by every communicator and stream. Because AlltoAll returns while its kernel is still running, another-stream launch can reuse or free the same stage/signal buffers; concurrent host calls also race the metadata. Store this state per communicator/execution context and serialize or event-order reuse and free.

test/kernel/klx/device_api.xpu
```diff
@@ -0,0 +1,588 @@
...
+                   (clusternum != g_a2a.clusternum);
+  if (ensureA2AScratch(firstTile, npes, clusternum) != 0)
+    return flagcxSystemError;
+  if (allocated) {
+    flagcxXshmemRendezvousKernel<<<1, 1, cs>>>(state);
+    xpu_wait(cs);
+  }
+
+  flagcxXshmemAlltoall1launchKernel<<<clusternum, corenum, cs>>>(
+      send, recv, g_a2a.stage, g_a2a.signal, (int)count, (int)tileElems,
```
MC952-arch: **Fix location: `test/kernel/klx/device_api.xpu:541-585`.**

`count` is accepted as `size_t` but cast to `int` here (and `tileElems` is also narrowed). Values above `INT_MAX` wrap and produce incorrect indexing/loop bounds. Chunk the operation as the AllReduce launcher does, or reject counts that cannot be represented before allocating or launching.

test/kernel/klx/device_api.xpu
```diff
@@ -0,0 +1,588 @@
...
+
+  long tileElems = xshmemTileElems1launch(npes);
+  long firstTile = ((long)count < tileElems) ? (long)count : tileElems;
+  bool allocated = (g_a2a.stage == nullptr) || (g_a2a.signal == nullptr) ||
+                   (npes != g_a2a.npes) || (firstTile > g_a2a.stageElems) ||
+                   (clusternum != g_a2a.clusternum);
+  if (ensureA2AScratch(firstTile, npes, clusternum) != 0)
+    return flagcxSystemError;
+  if (allocated) {
+    flagcxXshmemRendezvousKernel<<<1, 1, cs>>>(state);
```
MC952-arch: **Fix locations: `test/kernel/klx/device_api.xpu:329-370`, `test/kernel/klx/device_api.xpu:576-587`.**

The results of the XPU waits/launches are not checked; AlltoAll returns success immediately after dispatch. A bad state or invalid launch can therefore surface as `flagcxSuccess`. Check the runtime status and translate failures to a FlagCX error, as the CUDA implementation does with its launch error.

MC952-arch (changes requested):
The KunlunXin/XSHMEM path has merge-blocking build and correctness issues: the backend selector checks `USE_XSHMEM` although the repository propagates `USE_SHMEM`; the main XPU compile/device-link flags are incomplete; the KLX test dispatcher does not produce the artifacts or symbol surface required by its consumers; and both tiled collectives can reuse a remote ping-pong slot before the peer has consumed it. The lifecycle, scratch concurrency, large-count, and error-reporting issues below should also be resolved. No KunlunXin CI result was present for the reviewed head.

MC952-arch (approved):
LGTM

PR #559, opened 2026-08-27
[Others] Adapt new FlagCX interfaces for MetaX backend

zhujiaxiaowang:
PR Category
Others

PR Types
New Features

PR Description
[adaptor/metax] Implement adaptation for new FlagCX interfaces:
memAlloc
memFree
commRegister
commDeregister
commWindowRegister
commWindowDeregister
getHandleForAddressRange
hostGetDevicePointer
hostRegister
hostUnregister
symPhysAlloc
symPhysFree
symFlatMap
symFlatUnmap
symMulticastSupported
symMulticastCreate
symMulticastBind
symMulticastTeardown
symMulticastFree

### DIFF
--- a/flagcx/adaptor/device/maca_adaptor.cc
+++ b/flagcx/adaptor/device/maca_adaptor.cc
@@ -101,6 +101,14 @@ flagcxResult_t macaAdaptorGetVendor(char *vendor) {
   return flagcxSuccess;
 }
 
+flagcxResult_t macaAdaptorHostGetDevicePointer(void **pDevice, void *pHost) {
+  if (pDevice == NULL || pHost == NULL) {
+    return flagcxInvalidArgument;
+  }
+  DEVCHECK(mcHostGetDevicePointer(pDevice, pHost, 0));
+  return flagcxSuccess;
+}
+
 flagcxResult_t macaAdaptorGdrMemAlloc(void **ptr, size_t size,
                                       void *memHandle) {
   if (ptr == NULL) {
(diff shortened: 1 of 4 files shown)


flagcx/adaptor/device/maca_adaptor.cc
```diff
@@ -351,42 +359,338 @@ flagcxResult_t macaAdaptorStreamWriteValue64(flagcxStream_t, void *, uint64_t,
...
+    return flagcxInvalidArgument;
+
+  mcMemGenericAllocationHandle *mcHandle =
+      (mcMemGenericAllocationHandle *)malloc(
+          sizeof(mcMemGenericAllocationHandle));
+  if (mcHandle == NULL)
+    return flagcxSystemError;
+
+  // Retain the physical allocation handle from the VMM-backed pointer
+  DEVCHECK(mcMemRetainAllocationHandle(mcHandle, ptr));
```
MC952-arch, line 399: **Fix location: `flagcx/adaptor/device/maca_adaptor.cc:385-415`.**

`DEVCHECK` returns immediately, so this failure leaks the just-allocated `mcHandle`. The later `mcMemGetAddressRange` and export calls also return without releasing the successfully retained handle, and the short-buffer branch frees only the wrapper. Since `*physHandle` is assigned only on success, the caller cannot clean any of these resources. Please use explicit status checks and a common cleanup path that releases the retained handle when acquired and frees the wrapper on every failure.

flagcx/adaptor/device/maca_adaptor.cc
```diff
@@ -351,42 +359,338 @@ flagcxResult_t macaAdaptorStreamWriteValue64(flagcxStream_t, void *, uint64_t,
...
+      peerHandle = selfHandle;
+    } else {
+      int fd = *(int *)peerHandles[i];
+      DEVCHECK(
+          mcMemImportFromShareableHandle(&peerHandle, (void *)(uintptr_t)fd,
+                                         mcMemHandleTypePosixFileDescriptor));
+    }
+    mcDeviceptr_t slot =
+        (mcDeviceptr_t)((uintptr_t)base + (uint64_t)i * allocSize);
+    DEVCHECK(mcMemMap(slot, allocSize, 0, peerHandle, 0));
```
MC952-arch, line 466: **Fix location: `flagcx/adaptor/device/maca_adaptor.cc:429-474`.**

After the VA reservation succeeds, any import/map/access failure returns through `DEVCHECK` without undoing prior slot mappings, releasing the current imported handle, or freeing `base`. Because `*flatBase` is assigned only after the loop, the caller has no address with which to clean up. Track successfully mapped slots and unwind them plus the reservation on every error path.

flagcx/adaptor/device/maca_adaptor.cc
```diff
@@ -351,42 +359,338 @@ flagcxResult_t macaAdaptorStreamWriteValue64(flagcxStream_t, void *, uint64_t,
...
+  return flagcxSuccess;
+}
+
+flagcxResult_t macaAdaptorSymFlatUnmap(void *flatBase, size_t allocSize,
+                                       int nPeers) {
+  if (flatBase == NULL)
+    return flagcxSuccess;
+  mcDeviceptr_t base = (mcDeviceptr_t)flatBase;
+  size_t totalSize = allocSize * nPeers;
+  DEVCHECK(mcMemUnmap(base, totalSize));
```
MC952-arch, line 483: **Fix location: `flagcx/adaptor/device/maca_adaptor.cc:477-485`.**

Setup maps each peer slot separately with `mcMemMap(slot, allocSize, ...)`, but teardown attempts one combined `totalSize` unmap. CUDA-compatible VMM APIs require unmap ranges to correspond to mapped ranges, so this can fail and the immediate-return macro then skips `mcMemAddressFree`. Unmap each peer slot using the original `allocSize`, then release the full reserved VA range.

MC952-arch (changes requested):
The MetaX/MCCL API adaptation compiles and preserves the pre-2.30.4 fallback, but the newly enabled VMM path has three resource-lifecycle defects: physical handles leak on allocation/export errors, partial flat-map failures cannot be rolled back by the caller, and teardown does not mirror the per-peer mapping granularity. The current MetaX symmem CI also skips MPI execution and globally disables VMM, so these paths are not runtime-tested. Please fix the cleanup/teardown behavior and enable focused MetaX VMM coverage.

MC952-arch (approved):
LGTM

PR #561, opened 2026-08-28
[CICD] Replace the standalone Nexus upload with the shared build-infra workflow

### LINKED ISSUE #481: DEB Nexus upload is structurally broken (cross-workflow artifact round-trip fails on h20)
`upload-nexus.yml` has never succeeded — every tagged run has failed. The root causes are structural, not a Nexus availability problem:

1. **Cross-workflow download fails on the self-hosted runner.** `upload-nexus.yml` runs on `h20`, which can reach the internal Nexus (`resource.flagos.net`) but not `codeload.github.com` / `api.github.com` reliably. `dawidd6/action-download-artifact` times out after retries (e.g. runs `26384666396` for `v0.13.0-rc0.1`, `25789420296` for `v0.12.0`).
2. **Stale / expired artifacts.** When the download does succeed, it resolves to the *latest successful* `build-deb.yml` run, which can be an old one whose artifacts have already expired (retention is 1 day) → `no downloadable artifacts found (expired)` (e.g. run `23571814496` for `v0.11.0`).
3. **Build is disabled.** `Build Debian Packages` is currently `disabled_manually`, so tags don't even produce fresh DEB artifacts to upload.

### Proposed fix

Mirror the RPM approach (#476): publish DEBs to `flagos-apt-hosted` **directly from the `build-deb` job** on the same self-hosted runner, gated on `v*` tags, using the locally-built artifacts. This removes the cross-workflow GitHub round-trip entirely (h20 only needs to reach the internal Nexus, which works).

Concretely:
- [ ] Add a tag-gated "Publish to Nexus" step to `build-deb.yml` that uploads the local `.deb` files via `curl`.
- [ ] Re-enable the `Build Debian Packages` workflow.
- [ ] Retire or repurpose `upload-nexus.yml` (the standalone cross-workflow upload).

### Reference

RPM side already adopts this pattern in #476 (`build-rpm.yml` → `Publish ... RPMs to Nexus YUM repository`).

shiptux:
## Why

The in-repo `upload-nexus.yml` has never published a package (#481):

- it `PUT`s each deb to `<repo>/<filename>` — the Nexus apt-hosted format only accepts a multipart `POST` to the repository root, so every upload fails with **HTTP 405** (visible in the 2026-08-04 dispatch logs)
- it authenticates with `REGISTRY_USERNAME` / `CONTAINER_REGISTRY` instead of the org-level `NEXUS_TOKEN`
- it ignores the rpm artifacts entirely
- it exits 0 and prints a success summary even when every upload failed

The tag runs for `v0.13.0-rc2.post1` and `v0.13.0` are recorded as failures; `flagos-apt-hosted` holds no FlagCX package today.

## What

Replaces the file with a thin caller of the shared [`flagos-ai/build-infra/.github/workflows/upload-nexus.yml`](https://github.com/flagos-ai/build-infra/blob/main/.github/workflows/upload-nexus.yml) (the one libtriton_jit uses), which:

- skips pre-release tags (`rc`/`a`/`b`/`dev`) on `push`, always publishes on `workflow_dispatch`
- waits for `build-deb.yml` (required) and `build-rpm.yml` (optional) of the same commit and downloads **all** their artifacts (every backend / target in the matrix)
- uploads deb → `flagos-apt-hosted` (multipart POST), rpm → `flagos-yum-hosted`, with `NEXUS_TOKEN`
- fails the job if packages were found but none uploaded

The build workflows are untouched. After merge, `v0.13.0` can be published with a manual dispatch (rebuild + upload) without re-tagging.

Fixes #481. Part of FEP-0019 Wave 1 (flagos-ai/community#59).


MC952-arch (approved):
LGTM

PR #562, opened 2026-09-01
[CICD] Add Fedora 43 to the NVIDIA RPM build matrix

shiptux:
## What

Adds a `fedora43` target to the NVIDIA RPM chain — the first Fedora (and first non-EL, non-openEuler) build in the matrix, as a template for extending FlagOS packaging to more RPM distributions.

- `packaging/rpm/dockerfiles/Dockerfile.rpm.nvidia`: new `fedora43` preparation case and build branch
- `packaging/rpm/specs/flagcx.spec`: new `external_ccl_runtime` switch excluding only `libnccl.so.2` from generated dependencies
- `packaging/rpm/dockerfiles/build-rpm-common.sh`: pass the new switch through
- `packaging/rpm/build-flagcx-rpm.sh`: `RPM_DISTRO=fedora43` support
- `.github/workflows/build-rpm.yml`: new `nvidia-fedora43` matrix row (plain `fedora:43` base image, GitHub-hosted runner)
- `packaging/README.md`: document the target

## Design

Unlike openEuler, Fedora is covered by NVIDIA's native CUDA repository (`developer.download.nvidia.com/compute/cuda/repos/fedora43`), so this target resolves as much as possible natively instead of reusing the pip-wheel prefix assembly:

| | openEuler 24.03 row | Fedora 43 row |
|---|---|---|
| CUDA toolchain | pip wheels → `/opt/cuda12` | native RPMs (CUDA 13.2, `/usr/local/cuda-13.2`) |
| libcuda link stub | from base image | `cuda-driver-devel-13-2` + `LIBRARY_PATH` (same mechanism NVIDIA's own CUDA images use) |
| dist tag | `RPM_DIST_TAG=.oe2403` | `%{?dist}` expands to `.fc43` natively, nothing to pass |
| RPM deps on vendor libs | all excluded (`external_vendor_runtime`) | kept for CUDA (satisfiable from NVIDIA's repo); only `libnccl.so.2` excluded (`external_ccl_runtime`) |
| NCCL | pip wheel (cu12) | pip wheel (cu13, NCCL 2.31 — no Fedora RPM exists) |

NCCL 2.31 ≥ 2.29, so the NCCL device-API backend selection in `makefiles/nvidia.mk` stays on the CCL traits path.

## Verification (local)

Built with `RPM_DISTRO=fedora43 ./packaging/rpm/build-flagcx-rpm.sh nvidia`:

[...]

PR #563, opened 2026-09-01
[CICD] Raise the RPM libnccl floor to 2.27 for the window API

shiptux:
## What

Bump the NVIDIA RPM's `Requires: libnccl` floor from 2.10 to 2.27 and replace the outdated comment.

## Why

`nccl_adaptor.cc` compiles its symmetric-window support whenever the build-time NCCL headers are newer than 2.27.0 (`NCCL_VERSION_CODE` guard). The RPM build containers install current NCCL releases (2.31.x today), so the packaged `libflagcx.so.0` carries undefined references to `ncclCommWindowRegister` / `ncclCommWindowDeregister` — APIs introduced in NCCL 2.27. The declared 2.10 floor no longer matches what the built artifact needs.

Observed while validating the v0.13.0 openEuler RPM on an A100 host that ships NCCL 2.21.5:

- linking a consumer executable against the installed `libflagcx.so` fails with `undefined reference to ncclCommWindowRegister`;
- at runtime, lazy binding lets basic collectives work (a 2-rank `perf_allreduce` sweep passes), but any call into the symmetric-memory path would abort on the unresolved symbol.

The floor only applies when `external_vendor_runtime` is unset (e.g. the Rocky 8 matrix line); the openEuler/pip-CUDA line excludes vendor requires and is unaffected.

## Note for CI

Since the effective runtime floor tracks the build-time headers, the build containers' NCCL version is part of the package interface. Worth pinning the `nvidia-nccl-cu12` version in the RPM/DEB build images rather than always taking the latest — separately, NCCL 2.31.2 also SIGFPEs (integer divide-by-zero inside `ncclAllReduce`) on a driver-535 A100 where 2.21.5 works, so newest is not automatically safest for the test matrix.


MC952-arch (approved):
LGTM

shiptux:
Rebased onto current main after #562 landed. The two changes touch the same guard in `flagcx.spec`, so the resolution keeps both: #562's `external_ccl_runtime` switch alongside this PR's raised floor.

```spec
%if 0%{?external_vendor_runtime} == 0 && 0%{?external_ccl_runtime} == 0
Requires:       libnccl >= 2.27
%endif
```

Semantics are unchanged from either side: the requirement is still skipped when the deployment supplies its own vendor or CCL runtime, and the floor is 2.27 for the window API.

PR #570, opened 2026-09-08
place libflagcx.so into the flagcx wheel package

taozhiwei:
place libflagcx.so into the flagcx wheel package

### DIFF
--- a/plugin/torch/setup.py
+++ b/plugin/torch/setup.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import sys
 
 # Disable auto load flagcx when setup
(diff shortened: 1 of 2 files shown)


plugin/torch/setup.py
```diff
@@ -46,6 +47,19 @@
...
 libs = ["flagcx"]
 
+# Bundle libflagcx.so into the wheel under flagcx/lib/ so the extension can
+# resolve it at runtime via the $ORIGIN/lib rpath, independent of the build tree.
+flagcx_lib_src = os.path.join(repo_root, "build", "lib", "libflagcx.so")
+bundled_lib_dir = os.path.join(plugin_dir, "flagcx", "lib")
+os.makedirs(bundled_lib_dir, exist_ok=True)
+if os.path.isfile(flagcx_lib_src):
+    shutil.copy2(flagcx_lib_src, os.path.join(bundled_lib_dir, "libflagcx.so"))
+else:
```
MC952-arch, line 57: **Fix location: `plugin/torch/setup.py:52-61`.**

This branch leaves any existing `flagcx/lib/libflagcx.so` in place. Since that directory is now ignored and `package_data` always includes `lib/*.so`, a prior backend/version can be silently bundled after the current `build/lib/libflagcx.so` disappears; cached extension output or a system library can let the wheel finish. Stage the library as part of the build command, or clear the destination first and fail wheel creation when the current artifact is unavailable.

MC952-arch (changes requested):
The `$ORIGIN/lib` layout is coherent, but wheel generation can silently reuse a stale staged `libflagcx.so` when the current build artifact is absent. Please make native-library staging fail closed and add a wheel-level regression test that checks archive contents, the extension runtime path, and stale-artifact handling.

PR #576, opened 2026-09-10
fix(musa): link default device API backend

yswang777:
### PR Category

PAL (Portable Abstraction Layer)

### PR Types

Bug Fix

### PR Description

Include `flagcx/adaptor/device_api/default_dev_api_backend.cc` in the MUSA platform build.

The MUSA makefile previously left `PLATFORM_EXTRA_SRCS` empty. The device API declares `devApiBackend` as an external symbol, and the default backend source provides its definition. Without linking that source, the MUSA wheel could be built but failed during import/dynamic loading with an undefined `devApiBackend` symbol.

This patch is intentionally limited to `makefiles/musa.mk`; other platform builds and the public ABI are unchanged.

### Validation

- `git diff --check`: PASS
- symbol declaration/definition and MUSA source-list audit: PASS
- the same patch was used to build importable FlagCX wheels on both nodes of a MUSA S5000 TP16 HyV4 deployment
- the final deployment kept IB/RoCE enabled and passed stability, GPQA thinking 26/26, and two 4096/1024/C64 runs

The target-device result validates the build fix in the frozen adaptation environment. Current-main clean MUSA wheel CI is still requested from the repository's MUSA runner.


MC952-arch (approved):
LGTM

PR #582, opened 2026-09-12
[PAL] Link the default device API backend for the remaining platforms

tengqm:
`libflagcx.so` links with `devApiBackend` undefined on five platforms, failing only at load
time.

`flagcx/adaptor/flagcx_device.cc` is matched by the `flagcx/adaptor/*.cc` wildcard, so it is
always compiled, and it dereferences `devApiBackend`. Nothing else defines that symbol: the
Makefile never globs `flagcx/adaptor/device_api/`, so a platform gets one only if its `.mk`
names it in `PLATFORM_EXTRA_SRCS`. `iluvatar_corex.mk`, `sunrise.mk`, `cambricon.mk`,
`tsm.mk` and `amd.mk` assign it empty.

The link still succeeds because the Makefile passes `-shared` without `-Wl,--no-undefined`,
which permits unresolved symbols. Nothing reports it until a consumer dlopens the library
with `RTLD_NOW`.

#576 fixed exactly this for MUSA alone. Apply the same one-line default to the other five,
matching `kunlunxin.mk` and `ppu.mk`. A shared `device_api/*.cc` wildcard would not work:
`default_`, `nccl_`, `nvshmem_` and `xshmem_` are mutually exclusive, each defining
`devApiBackend`.


MC952-arch (approved):
LGTM

PR #583, opened 2026-09-12
refactor(adaptor): rename the iluvatar_corex build key to iluvatar

tengqm:
`iluvatar_corex` is the only adaptor key in the tree carrying an underscore,
and Debian/RPM package-name grammars reject `_`. Every consumer that derives a
package name from the adaptor family therefore has to sanitize it at that
boundary. The key is not a shipped interface yet, so rename it to plain
`iluvatar` rather than carry the sanitization forever.

Renamed: the `USE_ILUVATAR` make variable, `makefiles/iluvatar.mk`, the
`USE_ILUVATAR_ADAPTOR` macro, the adaptor header `iluvatar_adaptor.h`, and the
`ADAPTOR_MAP` / `ADAPTOR_TO_MAKE_FLAG` / platform-probe entries in
`plugin/torch/_build_config.py`.

The pre-0.13 spelling keeps working for one release, so existing command lines
and environments are not broken by the rename. All three entry points for the
old name are accepted and route to the new one, each emitting a deprecation
warning:

- `make USE_ILUVATAR_COREX=1` — an alias in the `Makefile` guarded on
  `$(origin)`. It is honored only when it comes from the command line or the
  environment, so the `?=` defaults below cannot mask it, and it never overrides
  an explicit `USE_ILUVATAR`.
- `FLAGCX_ADAPTOR=iluvatar_corex`, or the equivalent `--adaptor` argument,
  normalized by `_canonical_adaptor()` before validation. The old spelling is
  deliberately kept out of `VALID_ADAPTORS` so it stays out of the user-facing
  adaptor lists, including the guidance printed when detection fails.
- `USE_ILUVATAR_COREX=1` in the environment of a pip build, detected by the same
  normalization pass.

Deliberately not renamed: the runtime vendor identity `"ILUVATAR_COREX"`
reported by `ixcudaAdaptorGetVendor` and compared in `cluster.cc`, and the
`FLAGCX_VENDOR_ILUVATAR_COREX` enum in `flagcx.h`. Those are matched between
peers at runtime and index the `cost_model` tables, so renaming one side alone
would misclassify clusters with no compile error.


Makefile
```diff
@@ -6,7 +6,7 @@ BUILDDIR ?= $(abspath ./build)
 # set to 0 if not provided
 USE_NVIDIA ?= 0
 USE_ASCEND ?= 0
-USE_ILUVATAR_COREX ?= 0
+USE_ILUVATAR ?= 0
```
MC952-arch, line 19: Better to deprecate USE_ILUVATAR_COREX instead of completely removing it

tengqm: Agreed. We are not supposed to break existing users, even if we are renaming things internally.

tengqm: Added conversion and warning on line 6.

PR #584, opened 2026-09-12
docs: list the missing backends in getting_started

tengqm:
`docs/getting_started.md` lists the accepted `<backend>` values twice: as a
bullet list, and inline in the host-api perf-test example command. Both lists
lag the Makefile:

- the bullet list is missing `USE_SUNRISE` and `USE_PPU`
- the example command additionally omits `USE_AMD`

Add them, so the documented set matches the `USE_*` flags the Makefile accepts.


MC952-arch (approved):
LGTM

PR #585, opened 2026-09-12
ci: pin the clang-format the format hooks run

tengqm:
The two clang-format hooks are `language: system`, so they use whatever
`clang-format` is on PATH, while CI pins `clang-format==14.0.6` in
`.github/scripts/ci/format.sh`. CI is green because of that pin; a
contributor with a newer clang-format gets a different verdict for the
same unmodified file.

clang-format 21 changed how `IndentCaseLabels` is applied and now
un-indents case labels, which is the opposite of what `.clang-format`
requests here. On a pristine checkout 21.1.8 reports violations and the
in-place hook rewrites them (152 lines in `ixcuda_adaptor.cc`).

Pin both hooks to 14.0.6 through `language: python` +
`additional_dependencies`, matching the version CI installs, and let
`format.sh` install only pre-commit.

Verified with 21.1.8 still on PATH: both hooks pass over the whole tree
and modify nothing.


MC952-arch (approved):
LGTM

PR #586, opened 2026-09-12
[UIL] Fall back to a system FlagCX library instead of failing on FLAGCX_PATH

tengqm:
### What

`FLAGCXLibrary._find_default_library()` resolved the library through a fixed
chain and raised `FileNotFoundError` the moment `FLAGCX_PATH` was set but held
no `libflagcx.so`. That made a system-installed FlagCX unusable whenever
`FLAGCX_PATH` was exported: the package-directory and source-tree candidates
were never reached, and no candidate ever considered a system-wide install.

This changes the resolver to:

1. Keep searching when `FLAGCX_PATH` holds no library, instead of raising there.
2. Accept the versioned soname (`libflagcx.so.0`) under `FLAGCX_PATH`. A
   packaged (deb/rpm) install ships only that name, never the unversioned
   development symlink.
3. Resolve a system-wide install with `ctypes.util.find_library("flagcx")`
   before giving up, so a library installed into the loader's search path is
   found without any environment variable being set.
4. Report the failure with every location actually searched. The old message
   hardcoded `(FLAGCX_PATH not set)` in one line, which was wrong whenever the
   variable was set and simply unusable.

### Why

Packaging FlagCX as a `.deb` installs `libflagcx.so.0` into `/usr/lib` with no
build tree on the machine. With `FLAGCX_PATH` unset the library is not found at
all and consumers that treat a load failure as "no FlagCX" silently fall back
to a different backend; with it set, the load raises. Neither is a usable
outcome for a package install, and the fix belongs in the resolver rather than
in each consumer.

### Testing

Added `FlagcxDefaultLibraryTest` to `plugin/interservice/test_flagcx_wrapper.py`:
an unusable `FLAGCX_PATH` still reaches the package directory, the versioned
soname under `FLAGCX_PATH` is accepted, the system lookup is used as the last
resort, and the error message names every searched location. Existing
`FlagcxUniqueIdTest` cases are untouched.

    Ran 8 tests in 0.007s
    OK

Four of the five new cases fail against the current `main` (three with the
premature `FileNotFoundError`, one on the message assertion), so they do
constrain the change rather than restate it.

Note that `plugin/interservice/test_flagcx_wrapper.py` is not wired into CI —
`.github/scripts/ci/run_unit_test.sh` drives the C++ suites under
`test/unittest/` only. The result above comes from running the file directly:

    python3 -m unittest plugin.interservice.test_flagcx_wrapper

### Scope

Only the resolver in `plugin/interservice/flagcx_wrapper.py` changes. Callers
that hardcode a library path and never call `_find_default_library()` are not
helped by this and need their own change; that is left out here deliberately.


MC952-arch (approved):
LGTM

PR #588, opened 2026-09-12
[CICD] Pin the Hygon CI image to its pre-SHCA digest

tengqm:
**What:** pin the Hygon CI image to a digest instead of the `b079886-hygon-dev`
tag, in all three places that name it — `.github/configs/hygon.yml`,
`perf-test-hygon` and `torch-api-test-hygon`.

**Why:** that tag was re-pushed on 2026-09-11 with a layer that runs an MLX→SHCA
conversion removing the `hcoll` package, while the Open MPI built into the image
still lists `libhcoll.so.1` in `libmpi.so.40`'s `DT_NEEDED`. Every MPI test link
then fails on undefined `hcoll_*` symbols. The pinned digest is the last one
whose runs passed; it still ships `hcoll 4.8.3223`, resolved through the RUNPATH
recorded in `libmpi.so.40`.

**Verify:** the digest pulls on the Hygon node, carries `hcoll` and
`libhcoll.so.1`, and `load_platform_config.rb hygon` resolves it unchanged; the
three edited files parse as YAML.

This is a stopgap — drop the pin once the image carries a working HCOLL again.


MC952-arch (approved):
LGTM

PR #591, opened 2026-09-13
[CICD] Make the ascend RPM installable: filter CANN requires, tag the dist

shiptux:
## What

Two small fixes for the ascend RPM line:

1. **Filter the auto-generated `libascendcl` / `libhccl` requires for the ascend backend.** CANN is deployed by the host through the `.run` installer, never as RPMs, so those SONAME requires can never resolve — `dnf` refuses to install the package (`nothing provides libascendcl.so()(64bit)`) even on a host that has CANN fully deployed. The filter expresses the same contract `external_vendor_runtime` gives the NVIDIA libraries: deployment supplies the vendor runtime, the RPM tracks only the system libraries.
2. **Give the ascend build a distro tag.** The CANN base images are openEuler but define `%dist` as empty, so the built RPM had an untagged Release (`0.8.0-1` instead of `0.8.0-1.oe2403`). `Dockerfile.rpm.ascend` now reconstructs the tag from `/etc/os-release`, matching the NVIDIA openEuler line.

Also fixes the pre-existing `%changelog` ordering in the spec (the rc2.post1 / rc0.1 entries were swapped, which trips `rpmbuild`'s `not in descending chronological order` check).

## Validation

Built with the v0.13.0 source in the CANN 8.5.0 openEuler image with this spec:

- before: `rpm -qp --requires` shows `libascendcl.so()(64bit)`, `libhccl.so()(64bit)` → install fails on any repo;
- after: requires are system libraries only, `provides libflagcx.so.0` intact, artifact tagged `.oe2403`.

Note: runtime CANN availability on the target host is unchanged — this only stops the RPM from demanding a repo provider for libraries that are never packaged as RPMs.


MC952-arch (approved):
LGTM

shiptux:
Rebased onto current main. This branch had been built on top of #562, so it carried that commit as well; now that #562 is merged, the rebase drops the duplicate and the PR contains only its own change (the Ascend requires filtering and dist tag). No conflicts.

PR #600, opened 2026-09-16
[CICD] Enable T-Head PPU unit-test workflowAliyun pr

canghaiX:
### PR Category

- CICD

### PR Types

- New Features
- Test Case
- Bug Fixes

### PR Description

- Add a pull-request triggered unit-test workflow for T-Head PPU.
- Configure the PPU CI image, runner label, PPU devices, RDMA devices, IPC, and privileged container access.
- Enable PPU unit-test suites for adaptor, core, P2P, RMA, runner, service, and symmem.
- Run P2P and RMA tests through the ACCL transport.
- Skip the unsupported same-host ACCL GPU READ tests while keeping the remaining P2P coverage and RMA tests enabled.
- Extend platform configuration loading to support both legacy and current Docker configuration formats.
- Update the Hygon CI image configuration.

### Validation

- Verified the PPU shell script with `bash -n`.
- Verified the PPU YAML configuration.
- Confirmed that RMA remains enabled.
- Confirmed that PPU RDMA device mappings are configured.


.github/configs/ppu.yml
```diff
@@ -11,6 +11,8 @@ docker_volumes: []
 
 docker_options:
   - --network=host
+  - --ipc=host
+  - --privileged=true
```
MC952-arch, line 15: **Fix location: `.github/configs/ppu.yml:13-39`.**

Please revert the PPU container-option changes. Current main already runs PPU for pull requests, so this delta adds no new test capability; it does make PR-controlled scripts execute as root in a privileged container on the self-hosted PPU runner. Combined with host network/IPC and the device/sysfs mounts, that breaks the runner trust boundary. Any future need for privileged PPU execution should use a separately justified trusted-ref workflow.

.github/workflows/unit-test-ppu.yml
```diff
@@ -0,0 +1,21 @@
+name: T-Head PPU Unit Tests
+
+on:
+  pull_request:
```
MC952-arch: **Fix location: `.github/workflows/unit-test-ppu.yml:1-21`.**

Please delete this workflow. The existing unit-test.yml already triggers on pull requests, discovers ppu.yml in its all-platform matrix, and calls the same reusable workflow. The current check rollup shows two complete seven-suite PPU runs under separate concurrency groups, so this file only duplicates existing coverage.

.github/scripts/set_env/ppu.sh
```diff
@@ -60,9 +60,9 @@ flagcx_ci_configure_suite() {
       export FLAGCX_CI_EXPECT_NET_ADAPTOR=BAREX
       ;;
     p2p)
-      # These suites call the IBRC vtable directly. The Engine tests use the
-      # runtime transport selector and are retained for ACCL coverage.
-      export GTEST_FILTER="-P2pAdaptorStruct.*:P2pAdaptorTest.*:P2pLoopbackTest.*:P2pBatchStruct.*:P2pBatchTest.*:P2pEngineRpcIbTest.ConnectAcceptIsLocalSameHost"
+      # Keep ACCL engine coverage, but skip the same-host GPU READ tests until
+      # ACCL supports this PPU self-loopback path; RMA remains enabled separately.
+      export GTEST_FILTER="-P2pAdaptorStruct.*:P2pAdaptorTest.*:P2pLoopbackTest.*:P2pBatchStruct.*:P2pBatchTest.*:P2pEngineRpcIbTest.ConnectAcceptIsLocalSameHost:FlagcxP2pEngineReadTest.*"
```
MC952-arch, line 65: **Fix location: `.github/scripts/set_env/ppu.sh:63-65`.**

Please revert this exclusion. It disables the entire FlagcxP2pEngineReadTest fixture without adding replacement coverage, while the PR's PPU workflow is already redundant with the existing matrix workflow. If a concrete hardware limitation requires a skip, handle it in a focused follow-up with a narrow condition and an explicit restoration criterion.

.github/configs/hygon.yml
```diff
@@ -2,12 +2,7 @@
 hardware_name: hygon
 display_name: Hygon DCU Tests
 
-# Pinned by digest, not by the b079886-hygon-dev tag: the rebake pushed under that
-# tag on 2026-09-11 runs an MLX->SHCA conversion that removes the hcoll package,
-# while the Open MPI in the image still has libhcoll.so.1 in libmpi.so's DT_NEEDED
-# and RUNPATH. Every MPI link then fails on undefined hcoll symbols. Drop the pin
-# when the image carries a working HCOLL (or an Open MPI rebuilt without it) again.
-ci_image: harbor.baai.ac.cn/flagos-dev/flagcx@sha256:00db17623d118e5c9c7788799c67c36a430d4d43550fc0f42476001e353a4140
+ci_image: harbor.baai.ac.cn/flagos-dev/flagcx:hygon-dev-fixed-20260912
```
MC952-arch, line 5: **Fix locations: `.github/configs/hygon.yml:5`, `.github/workflows/test.yml:160-161`.**

This returns Hygon unit tests to a mutable tag even though the removed comment records that a previous tag rebake broke MPI linking. It also leaves the Hygon perf job pinned to the old digest while its comment still claims the two references match. Please resolve the fixed image to a digest and update both locations together, or explicitly document and test an intentional split.

.github/scripts/ci/load_platform_config.rb
```diff
@@ -26,24 +27,40 @@
...
   runner_labels
-  docker_volumes
-  docker_options
   set_env
   unit_test_suites
 ]
 missing = required_keys.reject { |key| config.key?(key) }
 abort "#{config_path}: missing required keys: #{missing.join(', ')}" unless missing.empty?
 
+unless config.key?("docker_volumes") || config.key?("container_volumes")
```
MC952-arch, line 36: **Fix location: `.github/scripts/ci/load_platform_config.rb:36-63`.**

No selectable in-tree platform config uses container_volumes or container_options at this head; all use docker_volumes/docker_options, and this loader only reads .github/configs. Please remove this unused compatibility path from the PR unless there is a separate current consumer that can be identified and tested.

MC952-arch (changes requested):
Needs Major Rework. The PPU portion should be removed rather than revised: current main already runs PPU through the all-platform PR matrix, so this PR adds no new PPU target or capability. It instead duplicates all seven suites, broadens PR code to privileged execution on a self-hosted runner, and excludes existing GPU READ tests. The legacy config parser also has no current consumer. Separately, the Hygon image change should remain digest-pinned and synchronized with the perf workflow.

PR #606, opened 2026-09-17
packaging(deb): sync the changelog before building, as the RPM path does

shiptux:
## Problem

`packaging/sync-changelog.py` regenerates both `debian/changelog` and the RPM spec's `%changelog` from `docs/CHANGELOG.md`, but only the **RPM** build script calls it (`packaging/rpm/build-flagcx-rpm.sh`). The Debian path never has, so `debian/changelog` is maintained by hand and drifts:

- `packaging/debian/changelog` currently starts at **0.13.0-rc0.1**
- the repository is already on **0.14.0-rc2.post1**

Since the deb package version is derived from that changelog, the drift is not cosmetic — it decides the version of the package users install.

## Change

Run the same sync step in `packaging/debian/build-helpers/build-flagcx.sh`, right before the container build, with the same non-fatal handling as the RPM script (a failure warns, it does not abort the build). Eight lines, no other behavior touched.

## Verification

Ran the script's new block in a clean tree: `debian/changelog` went from `0.13.0-rc0.1` to `0.14.0-rc2.post1` and the spec's `%changelog` was regenerated consistently, both matching `docs/CHANGELOG.md`. Reverted afterwards — this PR contains only the build-script change, so the regenerated changelog lands on the next real build.

## Note

This makes the two formats consistent **within FlagCX**. FlagCX is currently the only component repository with both this script and a `docs/CHANGELOG.md`; a proposal for making the mechanism available across FlagOS repositories will follow separately.


MC952-arch (approved):
LGTM

